PR #60, opened 2026-05-27
chore(ci): validate bumper version input before downstream use

dewet22:
## What

Adds a single regex validation guard near the top of `bump-givenergy-modbus.yml`'s "Determine target version and base branch" step. Rejects any `NEW` that doesn't match `^[0-9]+\.[0-9]+\.[0-9]+(-?[a-z0-9.]+)?$`, with `::error::` + `exit 1`.

## Why

`NEW` is sourced from `github.event.client_payload.version` / `inputs.version` and flows into several shell-interpolated contexts downstream:

| Site | Vector |
|---|---|
| `branch="bump/givenergy-modbus-${NEW}"` + `git checkout -B "$branch"` + `git push -f origin "$branch"` | Git ref injection (`:`, `^`, `..`) |
| `echo "new=$NEW" >> "$GITHUB_OUTPUT"` | `GITHUB_OUTPUT` injection via newline |
| Python `pattern.subn(rf'\g<1>{new}\g<2>', text)` in the bump step | Regex backref subversion via `\g<...>` |
| `git commit -m "...${NEW}"`, `gh pr create --title "..." --body "...@${NEW}](.../v${NEW})..."` | Log/markdown injection |

The trust boundary is the `BUMP_PAT` used to fire the `repository_dispatch` — small but non-zero attack surface (PAT compromise, supply-chain). A single validation guard at the top defends every downstream use rather than scattering escapes everywhere.

## Acceptance behaviour (verified locally)

**Accepts** every historical modbus release tag — `2.0.4`, `2.0.0rc1`, `2.0.0a6`, `2.0.0a3` — plus forward-looking PEP 440 forms (`.dev0`, `.post1`, `-alpha.1`).

**Rejects** empty string; every shell metacharacter (`;`, `\$()`, backticks, `|`, `&`); path traversal (`../etc`); newline injection; regex backref injection (`\g<0>`); quote injection; git ref operators (`:`, `^`, `..`); branch-style strings (`main`, `v2.0.0`).

## What's not in scope

- The two stale orphan branches on origin (`bump-python-floor-for-modbus-v2`, `claude/beautiful-chaplygin-c3f28d`) — separate cleanup.
- Pre-existing `dashboard/generate.py` ruff format drift on main — unrelated.

Refs: [GitHub Actions workflow injection vulnerabilities](https://github.blog/security/vulnerability-research/how-to-catch-github-actions-workflow-injections-before-attackers-do/)

## Summary by CodeRabbit

* **Chores**
  * Added early validation in the release workflow to enforce a PEP 440–like X.Y.Z (optional suffix) version format, emitting an error and halting the job for invalid versions to prevent malformed version strings from progressing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dewet22/givenergy-hass/pull/60?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/bump-givenergy-modbus.yml
+++ b/.github/workflows/bump-givenergy-modbus.yml
@@ -33,11 +33,16 @@ jobs:
           # GITHUB_OUTPUT, and a Python regex backref template downstream. The
           # trust boundary here is the BUMP_PAT used to fire the dispatch — a
           # compromised PAT shouldn't be able to inject shell metacharacters,
-          # newlines (GITHUB_OUTPUT injection), or `\g<...>` (regex backref
-          # subversion in the bump step). The character class permits every
-          # tag the modbus repo has historically published (`X.Y.Z`,
-          # `X.Y.ZrcN`, `X.Y.ZaN`) plus future PEP 440 `.dev`/`.post` suffixes.
-          if ! [[ "$NEW" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-?[a-z0-9.]+)?$ ]]; then
+          # newlines (GITHUB_OUTPUT injection), `\g<...>` (regex backref
+          # subversion in the bump step), or git ref operators (`..`, `^`, `:`
+          # in the branch name). The pattern accepts every modbus tag in
+          # history (`X.Y.Z`, `X.Y.ZrcN`, `X.Y.ZaN`) plus forward-looking
+          # PEP 440 `.dev`/`.post` suffixes; the inner repetition deliberately
+          # requires each dot-separated segment to be non-empty, so values
+          # like `2.0.0..` or `2.0.0.post..1` — which would pass a coarser
+          # `[a-z0-9.]+` class but then break `git checkout -B` downstream —
+          # are rejected here instead.
+          if ! [[ "$NEW" =~ ^[0-9]+\.[0-9]+\.[0-9]+([-.]?[a-z0-9]+(\.[a-z0-9]+)*)?$ ]]; then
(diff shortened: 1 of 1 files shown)


dewet22:
I found one edge case worth tightening before this lands: the new validation still allows consecutive dots in the suffix because `[a-z0-9.]+` accepts empty dot-separated segments.

For example, these pass the regex locally:

```text
2.0.0..
2.0.0.post..1
```

They then flow into `branch="bump/givenergy-modbus-${NEW}"`, where git rejects them as invalid branch names:

```text
fatal: 'bump/givenergy-modbus-2.0.0..' is not a valid branch name
```

That also means the current PR description overstates the coverage a little, since `..` is not fully rejected yet. I would either tighten the pattern so dot-separated suffix segments cannot be empty, or add an explicit `[[ "$NEW" == *..* ]]` rejection before writing to `GITHUB_OUTPUT`.

Everything else I checked in the workflow looked aligned with the intended hardening, and the GitHub checks are green.

dewet22:
Fixed in 43ea36f — tightened the suffix to `[-.]?[a-z0-9]+(\.[a-z0-9]+)*` so each dot-separated segment must contain at least one alphanumeric. The `2.0.0..` / `2.0.0.post..1` / trailing-`.` family all reject at validation time now, before `GITHUB_OUTPUT` is written or `git checkout -B` is invoked. The legitimate release-tag set (every modbus tag in history + PEP 440 `.dev`/`.post`/`-alpha.N`) is unchanged — verified locally against the same test corpus plus the two examples you flagged.

Also fleshed out the inline comment to call out the empty-segment carve-out explicitly so future-us doesn't loosen it again without thinking.

PR #62, opened 2026-05-27
feat: persist PlantCapabilities across restarts (#48)

### LINKED ISSUE #48: Persist PlantCapabilities across restarts to skip re-detection
## Background

`Client.detect()` probes the Modbus bus at startup to discover the plant topology — inverter type, number of LV batteries, HV stack addresses, meter addresses — and returns a `PlantCapabilities` object. The coordinator currently calls `detect()` on every HA startup, which adds latency and means the battery device count is re-probed each time.

Because discovery is the only source of truth for the battery count, there is also no user-accessible override: if detection gets the count wrong (e.g. a battery is slow to respond on the day HA restarts) the user has no way to correct it short of restarting HA and hoping detection succeeds next time.

## Proposed approach

Use HA's `Store` (the same mechanism used for the dashboard schema) to persist the serialised `PlantCapabilities` on first successful detection:

- On startup: load from store; if present and not stale, skip `detect()` and build `plant.capabilities` directly from the cached value
- On detection success: write the result back to the store
- Expose a way to force re-detection (e.g. a service call or a config-entry reload) for cases where the topology changes (battery added/removed)

This also provides an implicit override path: a sufficiently motivated user could edit the store file directly, though a UI option would be cleaner long-term.

## Prerequisite

**givenergy-modbus#72** — `PlantCapabilities` needs to be a Pydantic v2 model before it can be cleanly serialised/deserialised to JSON. The hass-side `Store` work should follow once that lands.

## Notes

- Tracking issue in givenergy-modbus for the Pydantic migration: dewet22/givenergy-modbus#72
- Related: dewet22/givenergy-modbus#43 (inverter-bound commands, same v1.1 batch)
- In the meantime a troubleshooting note has been added to the README explaining that battery count is auto-discovered and pointing users to open an issue with a frame capture if detection misfires

dewet22:
Closes #48.

## Summary

Caches the discovered plant topology in HA's `Store` and passes it to `Client.detect(prior=...)` on subsequent setup. On a topology match the library skips the full peripheral probe sweep (meters, LV batteries, BCU stacks) and just confirms the prior; on `PlantTopologyMismatch` the integration accepts the new topology, persists it, raises an advisory Repairs note, and triggers a config-entry reload so frozen-at-setup entity counts (notably batteries) reflect reality.

Library prerequisites landed in `givenergy-modbus 2.0.4` (already pinned): `PlantCapabilities.to_dict()` / `from_dict()` and `Client.detect(prior=...)`.

## Design notes

**Persistence shape follows HA Core's `Store` convention.** The library's `to_dict()` dict is the storage payload directly — no HA-side envelope, no in-payload `schema_version`, no `last_detected_at` timestamp. I went looking through ~12 HA Core integrations (`esphome`, `homekit_controller`, `homekit`, `google`, `bluetooth`, `thread`, `backup`, etc.) and that's the universal norm — `STORAGE_VERSION` on the `Store` constructor + `_async_migrate_func` on a subclass is the canonical migration mechanism, not a bespoke version field. Library-internal schema evolution is already handled by `from_dict()`.

**Cache I/O is inlined.** Three module-level helpers (`_capabilities_store`, `_load_capabilities`, `_save_capabilities`) co-located with the existing dashboard `Store` usage in `__init__.py`. No wrapper class — the helpers are ~25 LoC and the dashboard pattern in the same file inlines `Store` use directly, so this matches the codebase precedent.

**Topology mismatch UX**: auto-accept the new topology, persist it, schedule `async_reload` so entity counts catch up, and raise an `is_fixable=False` Repairs issue that's purely advisory (the reload already handled the fix).

**Force-redetect** via a new service `givenergy_local.redetect_plant` — clears the per-entry Store file and queues a reload.

**`_prior_capabilities` lives across in-process reconnects** — transient TCP drops re-enter `_connect()` and benefit from the same hint rather than running a cold sweep every time.

## Files changed

| File | Δ |
|---|---|
| `__init__.py` | +105 — helpers, topology callback, save-on-cold-start, redetect service handler |
| `coordinator.py` | +54/-6 — `prior_capabilities` + `on_topology_changed` constructor params, mismatch handling in `_connect()` |
| `const.py` | +1 — `SERVICE_REDETECT_PLANT` |
| `services.yaml`, `strings.json`, `translations/en.json` | +33 — user-facing strings |
| `tests/conftest.py` | +11 — real `PlantCapabilities` on `mock_plant` (was MagicMock attribute) |
| `tests/test_coordinator.py` | +110 — 4 new coordinator-level tests |
| `tests/test_init.py` | +332 — 12 new tests (4 helper units + 5 integration with patched helpers + 3 real-Store end-to-end + 2 service error branches) |

## Tests

99 pass total (was 83 baseline). The 16 new tests split:

**Coordinator-level (test_coordinator.py)**: prior threaded into `detect()`, cold path with `prior=None`, mismatch invokes callback and accepts actual, mismatch is safe without a callback.

**Cache helpers (test_init.py)**: hit / miss / library rejection / hand-edited TypeError / save shape.

**Integration (test_init.py)**:
- Patched-helper tests for redetect service, topology mismatch end-to-end, cold-start save.
- **Real-Store** tests with no helper patching — these are the gold-standard ones:
  - Warm-start cycle: cold setup writes through real `Store` → reload reads it back → prior passed to `detect()`.
  - Topology-mismatch reload cycle: seeded prior → mismatch → real save of actual → reload completes → second setup uses actual as prior.
  - Redetect service triggers cold reload via real `Store` (the reload's `detect()` is called with `prior=None`, which is only true if the file was really removed).
- Error branches: unknown device, device with no GivEnergy entry.

Coverage on all new code is complete. The 60 uncovered lines that remain in `__init__.py` are pre-existing handlers (`handle_reboot_inverter`, `handle_calibrate_battery_soc`, `handle_generate_dashboard`, `handle_capture_frames`, `_coordinator_for_device`) untouched by this PR.

## Out of scope

- Diagnostic sensor exposing cache age — would need a `last_detected_at` field that we're not currently persisting. Easy follow-up if useful.
- EMS topology caching for #52 — the same helper pattern extends naturally when that feature lands.

## Summary by CodeRabbit

* **New Features**
  * Cache and reuse plant topology per integration entry to speed startup and reduce detection
  * Added "Redetect Plant" service to clear cached topology and force rediscovery
  * Auto-accept, persist and reload when plant topology changes; creates a dismissible advisory issue

* **Bug Fixes**
  * Handle topology mismatches gracefully and persist actual capabilities

* **Documentation**
  * Add user-facing issue text and service translations

* **Tests**
  * Coverage for caching, redetect service, topology-mismatch, cold-start and store helpers

[Review Change Stack](https://app.coderabbit.ai/change-stack/dewet22/givenergy-hass/pull/62?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


custom_components/givenergy_local/__init__.py
```diff
@@ -36,6 +38,47 @@
...
+    None; we only add a guard around `from_dict()` for cases the library
+    rejects (hand-edited files, payloads from a future library schema, etc.).
+    Callers treat None as a cue to run a cold `detect()`.
+    """
+    payload = await _capabilities_store(hass, entry_id).async_load()
+    if payload is None:
+        return None
+    try:
+        return PlantCapabilities.from_dict(payload)
+    except (KeyError, ValueError, TypeError) as exc:
```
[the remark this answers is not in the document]
dewet22: Traced through `PlantCapabilities.from_dict()` against this. The library starts with `dict(data)`, so any non-dict payload (list/str/int) raises `TypeError` or `ValueError` at that first line — both already caught. I couldn't find a path inside `from_dict()` that would surface `AttributeError` on hand-edited JSON. If the library ever did start `AttributeError`-ing on non-dict input, that'd be a library-side contract bug I'd rather surface than silently absorb as a cache miss, so leaving the except tuple as-is.

custom_components/givenergy_local/__init__.py
```diff
@@ -87,16 +130,49 @@ async def async_migrate_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
...
+        ir.async_create_issue(
+            hass,
+            DOMAIN,
+            f"plant_topology_changed_{entry.entry_id}",
+            is_fixable=False,
+            is_persistent=False,
+            severity=ir.IssueSeverity.WARNING,
+            translation_key="plant_topology_changed",
+        )
+        hass.async_create_task(hass.config_entries.async_reload(entry.entry_id))
```
dewet22: I found one HA lifecycle detail worth tightening here. `ConfigEntries.async_reload()` now documents that integrations should prefer `async_schedule_reload()` when reloading from integration code, because the helper cancels pending setup retries before starting the reload task.

This manual `hass.async_create_task(hass.config_entries.async_reload(...))` pattern appears here and in the `redetect_plant` service. I would switch both to the helper so the behaviour matches HA’s intended reload path:

```python
hass.config_entries.async_schedule_reload(entry.entry_id)
```

and:

```python
hass.config_entries.async_schedule_reload(target_entry_id)
```

dewet22: Fixed in e17d26f — switched both call sites to `hass.config_entries.async_schedule_reload(entry_id)`. Confirmed the helper exists in HA Core 2026.5.x (`homeassistant/config_entries.py:2392`) and that HA Core uses it internally for the same purpose in five places (`async_get_or_create`, the device-registry reload paths, etc.). Existing tests still pass because the helper invokes `async_reload` via `hass.async_create_task` internally, so patched `async_reload` mocks still fire correctly.

PR #86, opened 2026-05-31
fix(coordinator): throttle the partial-refresh warning

### LINKED ISSUE #84: Quieten partial-refresh logging (debounce/aggregate the per-poll WARNING)
The partial-poll handling added in #125 logs a WARNING on **every** poll cycle that has a partial failure — `Partial refresh: N register read(s) failed; serving last-known values for those banks. Failures: [...]`. On a flaky or contended plant this floods the log (see Nick's logs on #52 — 8–12 occurrences over a couple of hours, repeating).

The behaviour itself is correct (serving last-known per-bank rather than dropping the whole plant), and the cumulative `partial_failures` counter + sensor already track it for diagnosis. So the per-cycle WARNING is largely redundant noise.

### Proposal
Downgrade / debounce the steady-state partial log:
- Log at WARNING only on the **first** occurrence (or transition), then periodically (every Nth cycle / every X minutes), relying on the `partial_failures` sensor for ongoing visibility — or move the steady-state line to DEBUG.
- Keep the structured `ReadFailure` detail (device addresses + bank) at whatever level we pick, so flaky-device diagnosis from logs still works.

### Why it matters
Partials are *expected* on some plants — the common triggers are device-level dongle garbage (givenergy-modbus#78) and multi-client contention (givenergy-modbus#132, e.g. running GivTCP alongside this integration). The logging shouldn't punish those users with a wall of WARNINGs for behaviour that's working as designed.

dewet22:
Fixes #84.

#125's partial-poll handling serves last-known values per-bank when a read times out (good), but logged a WARNING on **every** poll cycle that had a partial — flooding the log on flaky/contended plants (Nick's logs on #52, repeating each cycle).

Now:
- **First partial of a run** → WARNING (notes that further partials this run are at DEBUG)
- **Sustained run** → re-WARN every `_PARTIAL_LOG_EVERY` (20) polls
- **In between** → DEBUG
- **Clean poll** resets the run, so the next partial warns afresh

The cumulative `partial_failures` counter and the device-naming diagnostic sensor are unchanged — only the log cadence. Two tests cover the debounce-and-reset and the periodic re-warn. 151 tests pass; mypy-neutral.

## Release Notes

* **Improvements**
  * Enhanced logging efficiency for partial refresh operations. Warning logs are now intelligently throttled during sustained partial refresh sequences to reduce log verbosity while preserving comprehensive issue tracking. The system maintains periodic alert reminders to ensure visibility of ongoing problems and automatically resets throttling upon successful polling to restore full alerting responsiveness.

### DIFF
--- a/custom_components/givenergy_local/coordinator.py
+++ b/custom_components/givenergy_local/coordinator.py
@@ -35,6 +35,12 @@ _LOGGER = logging.getLogger(__name__)
 # so polling them every tick is wasteful.
 _FULL_REFRESH_INTERVAL = 300  # seconds (~5 minutes)
 
+# During a sustained run of partial polls, re-emit the partial warning at this
+# cadence (in polls) instead of every cycle. The first partial of a run always
+# warns; the in-between ones drop to DEBUG so a flaky/contended plant doesn't
+# flood the log. The cumulative partial_failures counter/sensor is unaffected.
+_PARTIAL_LOG_EVERY = 20
+
 
 class GivEnergyUpdateCoordinator(DataUpdateCoordinator[Plant]):
     """Wraps a long-lived Modbus Client, polling the inverter on a fixed interval.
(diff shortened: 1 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
Approved. I reviewed the current head and do not have any actionable concerns. I ran `uv run pytest -q` (150 passed), `uv lock --check`, and targeted ruff on the changed files; GitHub checks are green apart from CodeRabbit still being in progress at review time.

PR #100, opened 2026-06-02
fix: stop a slow BMS dropping a battery on detect; tidy service selectors

dewet22:
Two fixes from diagnosing a live report where a second battery (DZ2228G532)
silently disappeared and stayed gone across restarts.

## Root cause

`detect()`'s battery sweep reads pack #1 (0x32) with the robust `timeout/retries`
budget, but probes each *additional* pack (0x33+) with the library's stingy
defaults (`probe_timeout=0.5s`, `probe_retries=1`) and **breaks at the first
non-responding slot**. So a single transiently slow BMS during a reconnect
truncates the whole battery chain. That reduced one-battery topology then gets
persisted (a clean detect, no partial failure — so the #97 guard doesn't catch
it), and `detect(prior=…)` trusts it forever without re-probing the empty slot.

Confirmed live: the pack reported normally until a single clean transition at
the moment of an update's reconnect, then stayed unavailable for days; no
`plant_topology_changed` repair (the cold-detect persist path, not the mismatch
path). `redetect_plant` recovered it immediately, which validated the diagnosis.

## Fixes

1. **Widen the probe budget** at our `detect()` call to `probe_timeout=1.0s`,
   `probe_retries=3`. Because the sweep breaks at the first miss, a detect probes
   at most one empty slot, so the added latency is bounded to ~one slow probe per
   cold detect — cheap insurance against dropping a real pack. Also protects the
   warm/hinted path, where a transient miss would otherwise raise
   `PlantTopologyMismatch` and persist the reduced topology.

2. **Scope the service device selectors** for the six inverter-targeted services
   to `entity: domain: number` (the inverter/EMS controller has number entities,
   batteries don't), so battery devices no longer clutter those pickers.

## Notes

- This is the detect-robustness gap flagged as out-of-scope in the #97 plan.
- The library already exposed the probe knobs, so no modbus change is needed.
- Tests: the two exact-kwargs `detect()` assertions updated to expect the probe
  budget; the cold-start init test relaxed to assert `prior is None` without
  coupling to the constants. Full suite green (178), mypy at baseline.

## Summary by CodeRabbit

* **New Features**
  * Device detection now supports configurable timeout and retry settings for improved control during peripheral sweep operations.

* **Bug Fixes**
  * Service selectors now properly constrain device selection to display only relevant entity types.

* **Tests**
  * Updated test coverage to verify detection configuration parameters.

### DIFF
--- a/custom_components/givenergy_local/coordinator.py
+++ b/custom_components/givenergy_local/coordinator.py
@@ -41,6 +41,14 @@ _FULL_REFRESH_INTERVAL = 300  # seconds (~5 minutes)
 # flood the log. The cumulative partial_failures counter/sensor is unaffected.
 _PARTIAL_LOG_EVERY = 20
 
+# Per-slot probe budget for detect()'s peripheral sweep (batteries, meters).
+# More generous than the library defaults (0.5s / 1 retry): the battery sweep
+# breaks at the first non-responding slot, so a transiently slow BMS can
+# truncate the whole chain. The break means at most one empty slot is probed,
+# so the extra latency is bounded to ~one slow probe per detect.
+PROBE_TIMEOUT_SECONDS = 1.0
+PROBE_RETRIES = 3
+
 
 class GivEnergyUpdateCoordinator(DataUpdateCoordinator[Plant]):
     """Wraps a long-lived Modbus Client, polling the inverter on a fixed interval.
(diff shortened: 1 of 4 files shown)


dewet22-codex:

dewet22-codex (approved):
Approved. I reviewed the current head and do not have any actionable concerns. I ran `uv run pytest -q` (177 passed), `uv lock --check`, and targeted ruff on the changed files; the GitHub checks are also green.

PR #113, opened 2026-06-03
feat: rename inverter-output pair to PV Generation + modbus 2.1.2

dewet22:
## Summary

- Renames the two incorrectly-labelled "Inverter Output" sensors to **PV Generation Today** and **PV Generation Total** — IR44 and IR45/46 are PV generation registers, not inverter AC output (confirmed by givenergy-modbus #174/#176 sentinel cross-correlation).
- Both unique_ids migrate in place, preserving entity history.
- Pin bumped to `givenergy-modbus>=2.1.2`, which also makes `battery_max_power` return `None` (not `0`) for unknown DTCs — better signal for watt-limit consumers.
- Migration script and dashboard All-Time Totals retargeted to the new entity slugs.
- 4 new tests: value assertions + migration round-trip for both renamed entities.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected "Inverter Output" energy sensor labels to "PV Generation Today/Total" for accurate data representation.
  * Updated dashboard to display corrected PV generation metrics.

* **Improvements**
  * Added automatic migration for existing installations to seamlessly transition to renamed sensors while preserving entity history.

* **Chores**
  * Updated dependency requirements.

### DIFF
--- a/custom_components/givenergy_local/dashboard.py
+++ b/custom_components/givenergy_local/dashboard.py
@@ -285,7 +285,7 @@ def _energy_view(inv: str) -> str:
           - entity: {_i(inv, "grid_import_total")}
             name: Grid Imported
           - entity: {_i(inv, "pv_generation_total")}
-            name: Inverter Output
+            name: PV Generation Total
           - entity: {_i(inv, "charge_from_grid_total")}
             name: Charged from Grid
           - entity: {_i(inv, "battery_discharge_this_year")}
--- a/custom_components/givenergy_local/dashboard.py
+++ b/custom_components/givenergy_local/dashboard.py
@@ -284,7 +284,7 @@ def _energy_view(inv: str) -> str:
             name: Grid Exported
           - entity: {_i(inv, "grid_import_total")}
             name: Grid Imported
-          - entity: {_i(inv, "inverter_output_total")}
+          - entity: {_i(inv, "pv_generation_total")}
             name: Inverter Output
           - entity: {_i(inv, "charge_from_grid_total")}
             name: Charged from Grid
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -516,25 +516,25 @@ INVERTER_SENSORS: tuple[GivEnergyInverterSensorDescription, ...] = (
         value_fn=lambda inv: inv.e_ac_charge_today,
     ),
     GivEnergyInverterSensorDescription(
-        # NB: givenergy-modbus #174 found IR44 is PV generation, not inverter AC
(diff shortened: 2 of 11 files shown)


dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -516,25 +516,25 @@ class GivEnergyCoordinatorSensorDescription(SensorEntityDescription):
...
+        name="PV Generation Today",
         native_unit_of_measurement=UnitOfEnergy.KILO_WATT_HOUR,
         device_class=SensorDeviceClass.ENERGY,
         state_class=SensorStateClass.TOTAL_INCREASING,
         value_fn=lambda inv: inv.e_pv_generation_today,
     ),
     GivEnergyInverterSensorDescription(
-        key="e_inverter_out_total",
-        name="Inverter Output Total",
+        key="e_pv_generation_total",
```
dewet22-codex, line 532: This rename also needs to flow through `EXPOSE_RECOMMENDED_ENTITY_KEYS`. That curated list in `const.py` still contains `e_inverter_out_total`, but after this descriptor changes to `e_pv_generation_total` there is no entity with the old unique_id suffix, so `expose_recommended_entities` will silently skip this headline energy sensor. I’d update the constant/comment to the new PV-generation key and add a small assertion around the curated keys so this does not drift again.

custom_components/givenergy_local/dashboard.py
```diff
@@ -284,7 +284,7 @@ def col_series(entity: str, name: str, color: str) -> str:
             name: Grid Exported
           - entity: {_i(inv, "grid_import_total")}
             name: Grid Imported
-          - entity: {_i(inv, "inverter_output_total")}
+          - entity: {_i(inv, "pv_generation_total")}
             name: Inverter Output
```
dewet22-codex: This generated dashboard row now points at the PV-generation-total entity, but the display label is still “Inverter Output”. Since the PR is correcting that pair’s meaning, I’d rename the row to “PV Generated” or “PV Generation Total” here too, and mirror the same wording in `dashboard/template.yaml` so freshly generated dashboards do not keep the old mislabel.

custom_components/givenergy_local/sensor.py
dewet22: Fixed in 5069b94 — updated EXPOSE_RECOMMENDED_ENTITY_KEYS to e_pv_generation_total.

custom_components/givenergy_local/dashboard.py
dewet22: Fixed in 5069b94 — renamed the row label from "Inverter Output" to "PV Generation Total" in both dashboard.py and template.yaml.

PR #122, opened 2026-06-04
feat: dashboard PR B — battery OOS, AC output telemetry, battery maintenance mode

dewet22:
## Summary

Mid-priority observability gaps, building on PR A.

- **Battery Out Of Spec** (`binary_sensor`, #78) now surfaces on the Overview Status glance and Diagnostics → Faults & Warnings. Existing platform code didn't carry a `name=` in DeviceInfo, so HA was registering the entity_id as `binary_sensor.battery_out_of_spec` (no device prefix). Aligned its DeviceInfo with `sensor.py` (`name="GivEnergy Inverter <serial>"`) so the slug becomes `binary_sensor.givenergy_inverter_<serial>_battery_out_of_spec` — consistent with every other platform.
- **AC output Voltage / Frequency / Current** added to Diagnostics → Electrical alongside the existing AC input metrics (input/output labels disambiguated).
- **Battery Maintenance Mode** added to Diagnostics → Hardware & Firmware. Three-phase only; `skip_if_none` keeps the entity absent on single-phase inverters.
- `DASHBOARD_VERSION` → 9 so existing installs see the Repairs prompt.

Refs #78. Branched off `main`, independent from PR #121.

## Test plan

- [x] Full local test suite green (`uv run pytest`): 231 passed.
- [x] New tests assert battery_out_of_spec appears in both Status glance and Faults card; Electrical card carries AC output telemetry; Hardware card carries battery_maintenance_mode.
- [ ] Manual: existing install — verify `binary_sensor.battery_out_of_spec` keeps its entity_id (HA preserves it for existing installs); fresh install — verify the new prefixed entity_id is generated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #121, opened 2026-06-04
feat: dashboard PR A — Smart Load, AC-Coupled, mode controls, battery totals

dewet22:
## Summary

Closes the highest-priority gaps between entities the integration exposes and what the generated dashboard surfaces:

- **Smart Load slots 1–10** on Controls. Wired through `generate_dashboard(has_smart_load=…)`; defaults to True for inverter installs until `givenergy-modbus` exposes a capability we can gate on (modbus #181, targeted at 2.1.3). The Repairs prompt has shown for some users that the Smart Load section was sitting unused — it now has a home. Until the capability lands, plants without Smart Load hardware will see the rows render as 'unavailable'; the gating story tightens up once the modbus side ships.
- **AC-Coupled card** on Controls (Export Priority, EPS Enable, AC Charge Limit, AC Discharge Limit), gated by `capabilities.has_ac_config_block` from `PlantCapabilities` — present on AC-coupled / All-in-One, absent on hybrids.
- **Mode card additions**: Real Time Control switch + Inverter Max Output Active Power number (the active-power-rate knob added in #83).
- **Energy → All-Time Totals**: adds Battery Charged / Battery Discharged from the `e_battery_charge_total` / `e_battery_discharge_total` sensors (gated server-side by `skip_if_none`).
- **`DASHBOARD_VERSION` → 8** so the existing Repairs issue prompts users on existing installs to regenerate.
- Matching changes in `dashboard/template.yaml` for manual-setup users, with inline comments noting that the AC-Coupled and Smart Load cards should be removed by users whose hardware doesn't carry the corresponding capability.

Refs #52, #76, #83, #89, #90, #106.

## Test plan

- [x] Full local test suite green (`uv run pytest`): 235 passed (was 228 + 7 new).
- [x] New tests cover: Mode card extras present, AC-Coupled card hidden / emitted on capability, Smart Load 10-slot enumeration, suppressibility, EMS path excludes both new cards, Energy view lists battery totals.
- [ ] Manual: regenerate dashboard on a single-phase hybrid — AC-Coupled card absent, Smart Load present with rows unavailable.
- [ ] Manual: regenerate on an AC-coupled / All-in-One — AC-Coupled card present and entities resolve.
- [ ] Manual: regenerate on an EMS plant — neither new card appears.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Enhanced dashboard with new AC-Coupled controls for systems with AC battery capability
[...]

PR #123, opened 2026-06-04
feat: dashboard PR C — per-string PV, three-phase + EPS diagnostics, solar diverter

dewet22:
## Summary

Niche / completeness tail after PR A (#121) and PR B (#122). All entities here are gated server-side by `skip_if_none` or `single_phase_only`, so on plants without the underlying hardware the rows are absent (not "unavailable") — no UI noise for users who don't have these surfaces.

- **Diagnostics → PV Strings**: per-string daily energy (`pv_string_1_energy_today`, `pv_string_2_energy_today`).
- **Diagnostics → Electrical**: `grid_power_phase_1` (three-phase only), `backup_power`, `combined_generation_power`.
- **Energy → All-Time Totals**: `solar_diverter_energy_total` (hardware add-on).
- `DASHBOARD_VERSION` → 10.

`e_inverter_export_total` deliberately left out — wants a semantic confirmation against `grid_export_total` before either renaming, removing, or surfacing.

Branched off `main`, independent from PRs #121 and #122.

## Test plan

- [x] Full local test suite green: 231 passed.
- [x] New tests cover the three additions (PV Strings energy today, Electrical three-phase + EPS diagnostics, Energy solar diverter total).
- [ ] Manual: regenerate on a single-phase non-solar-diverter install — none of the new rows appear.

### DIFF
--- a/custom_components/givenergy_local/dashboard.py
+++ b/custom_components/givenergy_local/dashboard.py
@@ -9,7 +9,7 @@ import yaml
 # Increment whenever the generated YAML layout changes in a meaningful way.
 # __init__.py compares this against the last-generated version stored in HA's
 # persistent Store and raises a Repairs issue when they diverge.
-DASHBOARD_VERSION = 9
+DASHBOARD_VERSION = 10
 
 
 class _NoAliasDumper(yaml.SafeDumper):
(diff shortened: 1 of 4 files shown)


dewet22:
Gemini's finding was valid — the four new sensor descriptors (`p_grid_out_ph1`, `e_solar_diverter`, `p_backup`, `p_combined_generation`) were all missing `skip_if_none=True`. Fixed in e1c7322: the entities won't be created at all on installs where the underlying register returns `None`, so the dashboard will show "Entity not found" rather than "Unavailable/Unknown" on those installs. Same pattern as the other hardware-gated sensors already in the codebase.

The PR description's claim about server-side gating was optimistic — the fix makes it accurate.

PR #132, opened 2026-06-05
feat: live dashboard strategy (custom:givenergy) for #131

### LINKED ISSUE #131: Dashboard redesign — strategy + custom cards + mode-based UX
# Feature: dashboard redesign — strategy + custom cards + mode-based UX

## Summary

Replace the current `givenergy_local.generate_dashboard` "emit YAML to paste"
flow with a **dashboard strategy + custom card stack** that supports multiple
UX **modes**. The architecture must:

- Reproduce the existing six-tab dashboard faithfully (the `classic` mode is the
  v1 deliverable and the parity / fallback for every other mode).
- Resolve every entity from the registry (`unique_id` / `translation_key` +
  device), never via constructed entity-id strings — the existing static
  dashboard is currently broken because the inverter was assigned to an area
  and every entity_id was re-slugged with the `loft_` prefix.
- Treat **kiosk-mode + `panel: true`** as first-class design surfaces, not an
  afterthought, for the immersive modes (Flow / Glance / Story / Coach).
- Render correctly across **all viewports the HA Companion app and kiosk
  displays expose** — from ~380 px phone webviews up to 1920 px wall tablets —
  using container queries at the card level rather than viewport media
  queries.

## Motivation

The current generator produces a static YAML blob pasted into raw config. It's
frozen to the entity IDs that existed at generation time and silently rots when
anything changes.

**This is not hypothetical — it has already happened on a live install.** The
generated "Inverter" dashboard references `sensor.givenergy_inverter_sa2114g047_*`,
but the live registry now exposes `sensor.loft_givenergy_inverter_sa2114g047_*`
(area assignment re-slugged every entity_id). `ha_get_entity("sensor.givenergy_inverter_sa2114g047_pv_power")`
returns **Entity not found** — most cards on the existing dashboard are
currently dangling. A strategy resolving from the registry is structurally
immune to this.

Beyond the bug, the existing dashboard is hemmed in by Lovelace's card-grid
chrome and never really escapes the "compose small rectangles into bigger
rectangles" idiom. The redesign brief ([`dashboard-redesign-brief.md`](docs/design/dashboard-redesign-brief.md), mockup at
[`dashboard-redesign-exploration.html`](docs/design/dashboard-redesign-exploration.html) ·
  [rendered preview](https://htmlpreview.github.io/?https://raw.githubusercontent.com/dewet22/givenergy-hass/main/docs/design/dashboard-redesign-exploration.html)) explores five distinct UX directions
grounded in prior art (Tesla, GivEnergy portal, Victron VRM, SolarEdge,
Enphase, Sense, HA community packages), and concludes the dashboard should
ship as **selectable modes** rather than one consensus design.

## Architectural shift: strategy + custom cards + kiosk

The implementation surface is wider than just a Lovelace strategy:

1. **Dashboard strategy** (`custom:givenergy`) — generates the view structure
   based on the selected `mode` and the live registry. Also exposable as
   per-view strategies so users can mix-and-match modes across tabs.
2. **Custom Lit cards** — `givenergy-flow`, `givenergy-glance`,
   `givenergy-story`, `givenergy-coach` (and the existing
   `ge-cell-heatmap`). The chrome-free, full-viewport-respecting visuals live
   here, not in stock Lovelace cards. This is the same pattern as
   `power-flow-card-plus` (HACS) and proves the limits aren't HA's, they're
   stock card composition.
3. **Kiosk-mode integration** — the strategy emits per-view kiosk hints
   (`hide_header`, `hide_sidebar`) for immersive modes. Detection is
   first-class: if `kiosk-mode` isn't installed, the strategy omits the hints
   and the dashboard still works with the HA shell around it.
4. **`panel: true` views** for any mode whose card wants the entire viewport
   (Flow, Glance, Story, Coach in their immersive form). Analyst and Classic
   stay as standard multi-card views.

```yaml
strategy:
  type: custom:givenergy
  mode: classic         # classic | flow | glance | story | analyst | coach
  max_power_kw: 10
  serial: SA2114G047    # optional plant pin; default = all givenergy_local devices
```

Per-view override is supported too:

```yaml
views:
  - strategy: { type: custom:givenergy-glance }   # entry view: ambient
  - strategy: { type: custom:givenergy-analyst }  # second view: deep
  - strategy: { type: custom:givenergy-classic }  # third view: parity
```

## Existing dashboard to mirror (the `classic` mode)

`mode: classic` reproduces these six views. Card types in **bold** are HACS
custom dependencies — see "Custom card dependencies".

1. **Overview** — **power-flow-card-plus** · `glance` Status · `glance` Today ·
   **apexcharts-card** 24h power
2. **Energy** — **apexcharts-card** ×3 (30-day generation vs consumption; grid
   import vs export; battery charge vs discharge) · `entities` All-Time Totals
3. **Batteries** (sections, one per pack) — `gauge` SOC · `entities` Pack
   Details · `entities` Cell Temperatures · `entities` BMS Diagnostics
4. **Battery Health** (sections) — `markdown` · **ge-cell-heatmap** ·
   **apexcharts-card** per-cell voltages + temps (24h, with sample filtering) ·
   **apexcharts-card** battery power + pack SoC (24h)
5. **Controls** — `entities` Mode / Charging (+ slots) / Discharging (+ slots)
   / Smart Load (10 slots) / Maintenance (Redetect Plant, Sync Clock)
6. **Diagnostics** — `entities` Integration Health / Faults & Warnings /
   Temperatures / Electrical / PV Strings / Hardware & Firmware / Integration
   (Regenerate, Capture Debug Frames)

This is the v1 deliverable: registry-driven, working entity references,
identical layout. Everything below is the modes-roadmap that builds on it.

## Modes — five UX directions

Picked to span the space, not to converge on one design. Detailed treatment in
the redesign brief; one paragraph each here.

- **Flow** — manufacturer-app idiom. Animated power-flow SVG centrepiece, big
  SOC ring, today-totals strip, time-range pills. The consensus design across
  Tesla, GivEnergy, Victron, SolarEdge.
- **Glance** — calm-tech / ambient. Status sentence, three huge numbers, a
  strip of health pills. Ideal kiosk-on-fridge / phone-widget / watch face.
- **Story** — editorial / data-journalism. The day rendered as an annotated
  PV-curve timeline with event markers ("06:14 first photon", "13:08 cloud
  bank") plus source/sink distribution bars. Borrows from Tesla's impact
  framing but commits to it.
- **Analyst** — terminal-style, dense, monospace. 24h power overlay, per-cell
  voltage spread for every pack, source/sink ledger, diagnostic counters.
  Grafana-energy meets the existing Battery Health tab.
- **Coach** — actionable, tariff-aware. "Right now you're making money for
  the next four hours." Recommendations (run dishwasher; schedule EV around
  tomorrow's forecast), Octopus tariff strip, monthly goals. Depends on
  directional power sensors, a forecast feed, and tariff data — see Dependencies.

## Viewport considerations (each mode has a target range)

Not every mode works at every size. The HA Companion app renders dashboards in
a phone-sized webview; kiosk-mode renders them at full-display scale (often
1920×1080 or larger). Container queries on the card root, not viewport media
queries, are the right primitive — the same card config may appear in a wide
panel view *and* as a small card on another dashboard.

| Mode     | Min viewport      | Native at               | Strategy on under-min viewport                                     |
|----------|-------------------|-------------------------|---------------------------------------------------------------------|
| Glance   | 320 px            | All — mobile-first      | (none — Glance always works)                                        |
| Flow     | 480 px            | Tablet → kiosk          | Reflow flow diagram vertically (solar→home→battery→grid stacked)    |
| Coach    | 380 px            | Mobile-app / desktop    | Tariff strip compresses; recommendation cards stack                 |
| Story    | 600 px            | Tablet → desktop        | Replace unified SVG timeline with vertical event-card list          |
| Analyst  | 1024 px           | Desktop / kiosk only    | Render an "open in browser at desktop width" notice + Glance summary |
| Classic  | (existing chrome) | Anywhere with chrome    | Standard HA card stacking applies                                   |

Two implications for the cards:

- Every custom card **must** declare a `getCardSize()` and **must** render
  correctly inside both a `panel: true` view (full-viewport, no surrounding
  chrome assumptions) and a sized card slot (subject to grid sizing). The
  Glance card in particular needs to render elegantly at both 380×240 (phone
  widget) and 1920×1080 (kiosk tablet) — type scales need to be deliberate at
  both ends, not just one.
- The strategy should set **`panel: true` and emit kiosk-mode hints** for
  Flow / Glance / Story / Coach when used as a whole-view mode, and **omit
  both** when the same card is used as one card on a composed view. The card
  itself should handle both rendering contexts via container queries.

## Suggested rollout

Ship the architecture incrementally; do not block on the full mode set.

1. **v1 — `classic` mode + registry-driven resolution.** Strategy generates
   the existing six-tab dashboard with working entity references. Fixes the
   dangling-IDs bug. No new cards. This is the most valuable single change in
   the issue and is independently shippable.
2. **v2 — `flow` mode + `givenergy-flow` custom card.** Validates the
   custom-card path beyond `ge-cell-heatmap`, hits the manufacturer-app
   idiom users expect, and exercises kiosk + `panel: true` for the first
   time. Highest value-per-effort of the immersive modes.
3. **v3 — `glance` mode.** Cheapest custom card; broadens kiosk coverage to
   small screens / widgets. Likely most-used mode on Companion app.
4. **v4 — `analyst` mode.** Largely composed from existing ApexCharts +
   `ge-cell-heatmap`; modest new custom rendering for the source/sink ledger.
   Desktop/kiosk only — see viewport table.
5. **v5 — `coach` mode.** Last, because it depends on directional power
   sensors (see Dependency), a solar-forecast entity, and a tariff feed.
   Highest novelty but biggest dependency tree.
6. **`story` mode** — schedule against user demand; could equally well be a
   daily-digest notification rather than a tab.

## Design constraints for distribution cards

Where distribution cards (HA 2026.2+) appear inside any mode, they must respect:

1. **Only extensive quantities** (`power`, `energy`) — never intensive ones
   (`battery`/SOC, `temperature`, `voltage`, `frequency`).
2. **Split sources from sinks** — never mix flow directions in one bar.
3. **Distribution segments can't be negative** — signed single flow sensors
   are unusable directly (see Dependency).
4. **Skip buckets with < 2 members.**
5. **Group by device first.**
6. **Filter by `platform == "givenergy_local"`** so GivTCP duplicates don't leak.
7. **Classify by a static `translation_key -> role` manifest**, not entity-ID
   patterns.

## Entity resolution (the rule that fixes the rot)

Resolve every entity from the registry by **`unique_id` / `translation_key`
scoped to the device**, then read the entity's *current* `entity_id`. Never
construct or assume an entity_id string scheme — area assignment, user renames,
and multi-device installs all change the prefix. This is the structural fix for
the dangling-IDs bug.

## Custom card dependencies

| Card                | Source         | Used in modes        | Notes                              |
|---------------------|----------------|----------------------|------------------------------------|
| `power-flow-card-plus` | HACS        | classic              | Existing dependency                |
| `apexcharts-card`   | HACS           | classic, analyst     | Existing dependency                |
| `ge-cell-heatmap`   | this repo      | classic, analyst     | Existing custom card               |
| `givenergy-flow`    | this repo (new) | flow                 | New — v2 deliverable                |
| `givenergy-glance`  | this repo (new) | glance               | New — v3 deliverable                |
| `givenergy-coach`   | this repo (new) | coach                | New — v5 deliverable                |
| `givenergy-story`   | this repo (new) | story                | New — scheduled by demand          |
| `distribution` (built-in) | HA 2026.2+ | flow, story, analyst | Feature-detect; gauge fallback     |
| `kiosk-mode`        | HACS           | flow, glance, story, coach | Feature-detect; omit hints if absent |

Feature-detect each: if a required custom element isn't registered, degrade
gracefully (markdown placeholder, built-in fallback, omitted feature) rather
than emitting a broken card.

## Entity role mapping

Keys verified against the live install (suffixes after the area prefix):

```js
const ENERGY_SOURCES = ["pv_energy_today", "battery_discharge_today", "grid_import_today"];
const ENERGY_SINKS   = ["house_consumption_today", "battery_charge_today", "grid_export_today"];
const PV_STRINGS_E   = ["pv_string_1_energy_today", "pv_string_2_energy_today"];   // unsigned
const PV_STRINGS_P   = ["pv_string_1_power", "pv_string_2_power"];                 // unsigned
const POWER_SIGNED   = ["grid_power", "battery_power"];                             // needs directional split
const INTENSIVE_SKIP = ["battery_soc", "*_voltage", "*_temperature", "ac_frequency"]; // gauges/tiles only
```

Note `pv_energy_today` (DC) vs `pv_generation_today` (AC) — pick one
convention; in/out totals won't balance exactly, and two honest cards beats
one faked flow.

## Strategy skeleton

```js
class GivEnergyDashboardStrategy {
  static async generate(config, hass) {
    const mode = config.mode ?? "classic";
    const opts = { maxPowerKw: config.max_power_kw ?? 10, serial: config.serial };
    const caps = {
      distribution: supportsDistribution(hass.config.version),    // 2026.2+
      kiosk: customElements.get("kiosk-mode") != null,
      flowCard: customElements.get("givenergy-flow") != null,
      // ... feature-detect every HACS / repo-shipped dep
    };

    // resolve from the registry, never from entity-id strings
    const devices = givEnergyDevices(hass, opts.serial);          // platform == givenergy_local

    const builder = MODE_BUILDERS[mode] ?? MODE_BUILDERS.classic;
    return builder(hass, devices, opts, caps);
  }
}
customElements.define("ll-strategy-givenergy", GivEnergyDashboardStrategy);

// each mode is also registered as a view strategy: custom:givenergy-<mode>
```

Per-mode builder returns `{ views: [...] }` with `panel: true` and
kiosk-mode hints set for immersive modes when `caps.kiosk` is true.

## Dependency: directional power sensors

`coach` mode (and any instantaneous-power distribution card elsewhere) requires
non-negative directional power sensors. Today, `grid_power` and `battery_power`
are signed singles. Expose `grid_import_power` / `grid_export_power` and
`battery_charge_power` / `battery_discharge_power` integration-side — both for
this work and for the Energy dashboard. PV-string power is already unsigned.

## Version gating

Distribution card requires HA **2026.2+**. Gate that path on
`hass.config.version`; on older versions fall back to glance/apexcharts
equivalents so parity holds. Bump the `hacs.json` minimum HA version only when
a non-gated feature requires it.

## Acceptance criteria

**v1 — `classic` mode (must ship)**

- [ ] `strategy: { type: custom:givenergy, mode: classic }` renders all six
      views at parity with the current generator.
- [ ] All entities resolved from the registry; survives area reassignment and
      renames; the `loft_` prefix bug cannot recur.
- [ ] `max_power_kw` and optional `serial` honoured as strategy options.
- [ ] Multi-inverter / multi-battery installs produce per-device sections.
- [ ] `generate_dashboard` service retained as the static "eject" option.

**v2+ — per-mode (incremental)**

- [ ] Each mode emits valid Lovelace config; `panel: true` and kiosk-mode
      hints applied for Flow / Glance / Story / Coach when used as a whole
      view; omitted when the same card is composed alongside others.
- [ ] Each custom card renders correctly at its declared min viewport (see
      table); under-min viewports degrade per the strategy column.
- [ ] Card rendering driven by container queries on the card root, so the
      same card config renders correctly inside a `panel: true` view and
      inside a normal sized card slot.
- [ ] Every HACS / repo-shipped dep is feature-detected; missing deps degrade
      gracefully (no broken-card errors visible to the user).
- [ ] Distribution cards (where used) respect the constraints above; gated
      to 2026.2+ with fallbacks below that.

## Out of scope

- Theming / card-mod styling beyond what each custom card ships with.
- A daily-digest email/notification surface for Story (track separately).
- Multi-language localisation of recommendation copy in Coach.

## Prototype references

- 📄 [`dashboard-redesign-brief.md`](https://github.com/dewet22/givenergy-hass/blob/main/docs/design/dashboard-redesign-brief.md)
  — research synthesis (Tesla, GivEnergy, Victron, Enphase, SolarEdge, Sense, HA community), gap analysis, rationale per direction.
- 🎨 [`dashboard-redesign-exploration.html`](https://github.com/dewet22/givenergy-hass/blob/main/docs/design/dashboard-redesign-exploration.html)
  ([rendered preview](https://htmlpreview.github.io/?https://raw.githubusercontent.com/dewet22/givenergy-hass/main/docs/design/dashboard-redesign-exploration.html))
  — interactive five-direction mockup against the live install's real numbers.

dewet22:
Closes #131 (v1 — the `classic` strategy).

## What this adds

A Lovelace dashboard *strategy* (`custom:givenergy`) that builds the existing six-tab dashboard but resolves every entity from the live registry on each render, so it can't silently rot the way the static `generate_dashboard` YAML does. It resolves by `unique_id` (`{serial}_{key}`) rather than constructing entity_ids, which makes it structurally immune to the HA 2026.6 area-prefix re-slug (`sensor.loft_givenergy_…`) that was leaving most cards dangling on my own install after I moved the inverter into an area.

Usage — create a dashboard, open the raw config editor, and set:

```yaml
strategy:
  type: custom:givenergy
  mode: classic
  max_power_kw: 10     # optional
  serial: SA2114G047   # optional, multi-plant pin
```

## Notes

- The old standalone `ge-cell-heatmap.js` is merged into the strategy module, so one file now serves both `custom:givenergy` and `custom:ge-cell-heatmap`. Both gained an area-prefix-agnostic entity lookup (the heatmap was still constructing `entity_id`s directly and broke under the same re-slug); the strategy additionally skips disabled entities so their rows don't dangle.
- `generate_dashboard` stays as the editable static "eject" path — nothing changes there for existing users.
- Added a small vitest harness (mock `hass`) mirroring the Python `test_dashboard.py` discipline, plus a Python parity guard that pins every key the strategy references to a real `EntityDescription.key`, so a renamed key fails loudly here rather than dangling silently. New `js tests` CI job runs the vitest suite.

## Known limitation

On a *hard* refresh (cache bypass) the strategy can occasionally hit HA's "Timeout waiting for strategy element" — a 5-second registration window that a cold module re-fetch can lose when queued behind other custom-card resources. This affects all network-loaded HA strategies, not just this one; a normal reload serves from cache and isn't affected. Documented in the README. I didn't want to ship the fragile workaround (registering an inline Lovelace resource by poking the resource collection) for an edge case that only bites manual hard-reloads.

## Testing

- 257 Python + 15 JS tests green; ruff/mypy/codespell clean.
- Verified live on my own install (2-battery plant, inverter in an area so entity_ids carry the `loft_` prefix): all six views render and resolve correctly.

## Summary by CodeRabbit

* **New Features**
  * Introduced a custom dashboard strategy (`custom:givenergy`) that dynamically generates Lovelace dashboards from live Home Assistant data, automatically resolving entities and persisting across renames.
  * Added cell-heatmap visualization for battery cell-balance monitoring.

* **Documentation**
  * Added configuration guide for the new dashboard strategy with classic layout support.

* **Chores**
  * Enhanced testing infrastructure with automated CI/CD workflow and test suite setup.


dewet22-codex:

custom_components/givenergy_local/www/ge-strategy.js
```diff
@@ -0,0 +1,1060 @@
...
+      });
+    }
+    if (!target) target = inverters[0] || null;
+
+    // batteries belonging to the target inverter (by via_device); fall back to
+    // all batteries when via_device isn't populated.
+    var ownBatteries = batteries.filter(function (b) {
+      return b.viaDeviceId === (target && target.deviceId);
+    });
+    if (!ownBatteries.length) ownBatteries = batteries;
```
dewet22-codex: This fallback also fires when the selected target legitimately has no batteries. In a multi-plant install where another inverter does have batteries with valid `via_device_id` links, the inverter-only dashboard will then display every battery from the other plant. I’d fall back to all batteries only when the registry provides no usable `via_device_id` relationships at all; otherwise an empty match should remain empty.

custom_components/givenergy_local/www/ge-strategy.js
```diff
@@ -0,0 +1,1060 @@
...
+    });
+
+    // pick the target plant: serial pin > sole inverter > first
+    var target = null;
+    if (opts.serial) {
+      target = inverters.find(function (p) {
+        return String(p.serial).toUpperCase() === String(opts.serial).toUpperCase();
+      });
+    }
+    if (!target) target = inverters[0] || null;
```
dewet22-codex: When an explicit `serial` pin does not match, this silently falls back to the first inverter. On a multi-plant install, a typo or stale dashboard config would therefore show another plant and make the Maintenance buttons target that plant instead. I’d only apply the first-inverter default when no serial was supplied; an unresolved explicit pin can return the friendly “No GivEnergy plant” view (ideally naming the missing serial) instead.

custom_components/givenergy_local/www/ge-strategy.js
```diff
@@ -0,0 +1,1060 @@
...
+      geDevices.set(d.id, { serial: ident[1], viaDeviceId: d.via_device_id || null });
+    }
+
+    // key -> entity_id per device, by stripping the device's `{serial}_` prefix
+    // off each entity's unique_id.
+    var byDevice = new Map();
+    for (var j = 0; j < entities.length; j++) {
+      var e = entities[j];
+      if (e.platform !== DOMAIN) continue;
+      if (e.disabled_by) continue; // skip disabled entities; they have no state
```
dewet22-codex: Because disabled entities are discarded before `classify()`, disabling the one marker entity used for a device type makes the whole device disappear from the strategy: disabling `p_pv` hides an inverter, and disabling `ems_plant_enable` hides an EMS, even if all their other entities remain enabled. I’d retain a separate set of all registered keys for classification while keeping disabled entries out of the renderable key-to-entity map.

custom_components/givenergy_local/www/ge-strategy.js
dewet22: Good catch — fixed. The fall-back to all batteries now only fires when the registry exposes no `via_device_id` links at all; when links exist, an empty match stays empty, so an inverter-only plant no longer borrows another plant's packs. Added a multi-plant regression test.

custom_components/givenergy_local/www/ge-strategy.js
dewet22: Fixed — the first-inverter default now only applies when no serial is supplied. An unresolved explicit pin leaves the target unset and shows the no-plant notice, naming the missing serial. Covered by a new test.

custom_components/givenergy_local/www/ge-strategy.js
dewet22: Right — fixed. Classification now runs against a separate set of all registered keys (including disabled ones), while the renderable key→entity_id map still excludes disabled entities. So disabling `p_pv` or `ems_plant_enable` no longer hides the device. New test disables `p_pv` and asserts the inverter still renders.

custom_components/givenergy_local/www/ge-strategy.js
```diff
@@ -0,0 +1,1060 @@
...
+    return "other";
+  }
+
+  // Resolve the plant topology and a key->entity_id map per device from the
+  // registry. Returns { plants: [...], byDevice: Map<deviceId, Map<key, eid>> }.
+  async function buildPlant(hass, opts) {
+    var res = await Promise.all([
+      hass.callWS({ type: "config/entity_registry/list" }),
+      hass.callWS({ type: "config/device_registry/list" }),
+    ]);
```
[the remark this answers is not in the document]
dewet22: I've wrapped the two `callWS` calls in try/catch and return a friendly "could not read the registry" card on failure — a worthwhile guard against a transient connection drop. One correction: `config/entity_registry/list` and `config/device_registry/list` aren't admin-gated in HA core (neither handler carries `@require_admin`; only the update/remove handlers do), so a non-admin viewer can render the dashboard fine — the catch is for connection failures rather than permissions.

custom_components/givenergy_local/www/ge-strategy.js
```diff
@@ -0,0 +1,1060 @@
...
+          if (!cfg || !Array.isArray(cfg.batteries) || !cfg.batteries.length)
+            throw new Error("ge-cell-heatmap: 'batteries: [serial, ...]' is required");
+          this._cfg = cfg;
+        }
+        set hass(hass) { this._hass = hass; this._render(); }
+        getCardSize() { return (this._cfg && this._cfg.batteries.length || 1) + 1; }
+
+        _render() {
+          const cfg = this._cfg, hass = this._hass;
+          if (!hass) return;
```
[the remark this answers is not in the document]
dewet22: Added the guard. It's in the browser-only heatmap card so it sits outside the Node test harness, but it's a cheap safety net for early-bootstrap renders.

custom_components/givenergy_local/www/ge-strategy.js
```diff
@@ -0,0 +1,1060 @@
...
+          // state so the heatmap resolves on area-assigned installs.
+          const fallback = {};
+          for (const [eid, st] of Object.entries(hass.states)) {
+            const i = eid.indexOf("givenergy_battery_");
+            if (i > 0) fallback[`sensor.${eid.slice(i)}`] = st;
+          }
+          const cellState = (s, n) => {
+            const id = `sensor.givenergy_battery_${s.toLowerCase()}_cell_${n}_voltage`;
+            return hass.states[id] || fallback[id];
+          };
```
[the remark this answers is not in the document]
dewet22: Done — the heatmap now resolves each canonical cell id to its actual (area-prefixed) id once and caches it per-instance, scanning only on a cache miss (which self-heals if an entity_id later changes). Avoids the full state scan on every render.

README.md
```diff
@@ -227,6 +227,24 @@ If the dashboard schema is updated in a future release, the integration raises a
...
 
+### Dashboard strategy (live, self-maintaining)
+
+To avoid that snapshot problem entirely, there's also a dashboard *strategy* that builds the same six-tab layout but resolves every entity from the registry each time the dashboard loads — so it doesn't go stale when a device moves area or an entity is renamed. I added it because the static YAML kept silently rotting on my own install after area reassignments. To use it, create a new dashboard, open the **raw configuration editor**, and set the whole config to:
+
+```yaml
+strategy:
+  type: custom:givenergy
+  mode: classic        # the only mode in this release
+  max_power_kw: 10     # optional; Overview 24h chart y-axis envelope (kW)
```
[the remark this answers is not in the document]
dewet22: Added `default 10` to the example. I've deliberately left the 1–100 range off: that bound lives on the `generate_dashboard` service schema, not the strategy — the strategy reads `max_power_kw` straight from the dashboard config with no validation, so it accepts any value. It's only the chart's y-axis envelope in kW.

dewet22-codex:

dewet22-codex (approved):
Approved. I re-reviewed the current head and do not have any actionable concerns. The three earlier registry-resolution findings are addressed with targeted regression coverage, and the current GitHub checks are green.

PR #137, opened 2026-06-06
feat: add glance mode (mode: glance) for the givenergy dashboard strategy (#131)

dewet22:
## Summary

Closes part of #131 — introduces `mode: glance` as a third selectable mode for the `custom:givenergy` dashboard strategy.

**What's new**

- `mode: glance` prepends a full-viewport Glance panel (`panel: true`) to the existing classic view set.
- `custom:givenergy-glance` — a new vanilla custom element (no build step) that renders:
  - A natural-language status sentence with a pulsing amber/green indicator dot. Eight states are derived from live sign-combinations of grid/battery/solar (e.g. "Solar covering the house and charging the battery, exporting the surplus." / "Battery powering the house." / "Drawing from the grid."). The dot turns amber when importing from the grid or when battery SOC drops below 20%.
  - Three large Fraunces numerals: Solar today (kWh), Battery SOC (%), House today (kWh). Each has a contextual sub-line: string breakdown or current kW for solar, per-pack SOC or charge/discharge direction for battery, import/export direction for house.
  - Health pills: battery count, per-pack SOC when multiple packs are present (amber at < 20%), imported/exported kWh today, active per-string generation.
- EMS plants fall back to the classic view set (no glance panel), matching the `flow` mode behaviour.
- Kiosk-mode hints (`hide_header`, `hide_sidebar`) are feature-detected and omitted when the integration is absent.
- Container-query responsive layout; signature guard skips DOM rebuilds when no referenced state has changed.
- `_STRATEGY_VERSION` bumped to `6` for cache-busting.

**Tests**

Four new JS tests (29 total): view set shape, EMS fallback, registry resolution under the `loft_` area prefix, and missing-entity omission.

## Test plan

- [ ] Add `strategy: { type: custom:givenergy, mode: glance }` to a dashboard.
- [ ] Confirm Glance panel is the first view; classic tabs follow.
- [ ] Status sentence matches the current system state (solar generating, battery charging/discharging, grid import/export).
- [ ] Dot is green when self-sufficient/exporting, amber when importing or SOC < 20%.
- [ ] Three big numbers populated (Solar today / Battery % / House today).
- [ ] Health pills show battery count, import/export today, and per-string when active.
- [ ] On a narrow viewport, the three numbers stack to a single column.
- [ ] EMS plant: `mode: glance` falls back to the classic EMS view set.

## Summary by CodeRabbit

* **New Features**
  * Added a new "glance" dashboard mode with an at-a-glance view displaying solar generation, battery state-of-charge, and household usage.
  * Includes visual indicators for battery availability, import/export totals, and per-string solar generation.

* **Chores**
  * Updated dashboard version to ensure clients fetch the latest configuration.

### DIFF
--- a/custom_components/givenergy_local/www/ge-strategy.js
+++ b/custom_components/givenergy_local/www/ge-strategy.js
@@ -385,6 +385,15 @@
     return view;
   }
 
+  // mode: all -- dev convenience: all mode panels in front of classic tabs.
+  // Not intended as a permanent user-facing mode; remove when modes split into
+  // separate dashboards.
+  function allViews(plant, opts) {
+    var a = makeAccessors(plant);
+    if (plant.target && plant.target.isEms) return classicViews(plant, opts);
+    return [glanceView(plant, a, opts), flowView(plant, a, opts)].concat(classicViews(plant, opts));
+  }
+
   function overviewView(plant, a, opts) {
     var cap = (opts.maxPowerKw || 10) * 1000;
     var cards = [];
@@ -1079,6 +1088,7 @@
     var views =
       opts.mode === "flow"   ? flowViews(plant, opts)   :
       opts.mode === "glance" ? glanceViews(plant, opts) :
+      opts.mode === "all"    ? allViews(plant, opts)    :
                                classicViews(plant, opts);
     return { title: "GivEnergy", views: views };
   }
--- a/custom_components/givenergy_local/www/ge-strategy.js
+++ b/custom_components/givenergy_local/www/ge-strategy.js
@@ -1579,25 +1579,44 @@
           var importing   = grid  != null && grid  < -THRESH;
 
           // Natural-language status sentence (ASCII-only).
+          // Structure: check net grid direction (exporting/importing/idle) first
+          // within the solarOn group so the sentence is correct even when solar
+          // is present but insufficient to cover demand or the source of battery
+          // charge is ambiguous.
           var sentence;
-          if (solarOn && charging && exporting)
-            sentence = "Solar covering the house and charging the battery, exporting the surplus.";
-          else if (solarOn && charging)
-            sentence = "Solar covering the house and charging the battery.";
-          else if (solarOn && exporting)
-            sentence = "Solar ahead of demand - exporting to the grid.";
-          else if (solarOn && discharging)
-            sentence = "Solar and battery covering the house.";
-          else if (solarOn)
-            sentence = "Solar covering the house.";
-          else if (discharging && !importing)
+          if (solarOn) {
+            if (exporting) {
+              if (charging)
+                sentence = "Solar covering the house and charging the battery, exporting the surplus.";
+              else if (discharging)
+                sentence = "Solar and battery exporting to the grid.";
+              else
+                sentence = "Solar ahead of demand - exporting to the grid.";
+            } else if (importing) {
+              if (charging)
+                sentence = "Solar and grid supplying the house and charging the battery.";
+              else if (discharging)
+                sentence = "Solar, battery, and grid supplying the house.";
+              else
+                sentence = "Solar and grid supplying the house.";
+            } else {
+              // Grid approximately idle: solar balanced against load and battery.
+              if (charging)
+                sentence = "Solar covering the house and charging the battery.";
+              else if (discharging)
+                sentence = "Solar and battery covering the house.";
+              else
+                sentence = "Solar covering the house.";
+            }
+          } else if (discharging && !importing) {
             sentence = "Battery powering the house.";
-          else if (discharging && importing)
+          } else if (discharging && importing) {
             sentence = "Battery and grid supplying the house.";
-          else if (importing)
+          } else if (importing) {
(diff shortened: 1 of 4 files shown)


dewet22-codex:

custom_components/givenergy_local/www/ge-strategy.js
```diff
@@ -1455,6 +1510,223 @@
...
+          var charging    = batt  != null && batt  < -THRESH;
+          var discharging = batt  != null && batt  >  THRESH;
+          var exporting   = grid  != null && grid  >  THRESH;
+          var importing   = grid  != null && grid  < -THRESH;
+
+          // Natural-language status sentence (ASCII-only).
+          var sentence;
+          if (solarOn && charging && exporting)
+            sentence = "Solar covering the house and charging the battery, exporting the surplus.";
+          else if (solarOn && charging)
```
dewet22-codex: These branches infer the source and sufficiency of power from independent sign flags, so several normal states produce a confidently incorrect sentence. For example, `solar=200`, `grid=-1500`, and an idle battery reaches the later `solarOn` branch and says “Solar covering the house” despite substantial grid import; grid charging reaches this branch and says solar is charging the battery; battery-backed export can be described as solar export. The measured load and power balance can distinguish at least the covering/import cases, while source attribution may need more cautious wording where the net sensors are ambiguous. I think the status logic needs to account for importing/exporting before claiming that solar covers demand or supplies another destination.

custom_components/givenergy_local/www/ge-strategy.js
```diff
@@ -1024,7 +1076,10 @@
       };
     }
     // Mode dispatch. Unknown/absent mode falls back to classic.
-    var views = opts.mode === "flow" ? flowViews(plant, opts) : classicViews(plant, opts);
+    var views =
+      opts.mode === "flow"   ? flowViews(plant, opts)   :
+      opts.mode === "glance" ? glanceViews(plant, opts) :
```
dewet22-codex: This makes `mode: glance` a public selectable mode, but the README example still lists only `classic | flow`, and the dashboard-strategy section explicitly says that `glance` is still to come. That leaves users following the documented setup unable to discover the new mode and gives contradictory release documentation. I think this PR needs the same short mode section and option-list update that accompanied `flow`.

custom_components/givenergy_local/www/ge-strategy.js
dewet22: Fixed: restructured the sentence logic to check the net grid direction (exporting/importing/idle) within the solarOn group first, so solar=200 + grid=-1500 now correctly produces "Solar and grid supplying the house." rather than claiming solar covers the house.

custom_components/givenergy_local/www/ge-strategy.js
dewet22: Fixed: updated the mode comment to `classic (default) | flow | glance`, removed glance from the "still to come" sentence, and added a `#### mode: glance` section to the README matching the structure of the flow entry.

dewet22-codex:

custom_components/givenergy_local/www/ge-strategy.js
```diff
@@ -1024,7 +1085,11 @@
...
+      opts.mode === "all"    ? allViews(plant, opts)    :
```
dewet22-codex: This exposes `mode: all` through the production strategy even though the accompanying comment and commit describe it as temporary development convenience that is intended to be removed before release. Merging this PR puts it on `main`, where any dashboard config can select it and it becomes part of the integration’s observable mode surface despite being omitted from the documented option list. I think the dev-only dispatch and tests need to come back out before merge, or the mode needs to be treated and documented as a supported feature rather than relying on a later removal step.

dewet22:
Thanks for the three-point review. Quick status on each:

**Status sentence logic** — the concern was already addressed by the fix commits (`33e4da2`, `a26f0b7`) that landed after the first review pass. The dispatch now checks `exporting`/`importing` before reaching any solar-only clause, so `solarOn=true, importing=true` always produces "Solar and grid supplying the house" rather than the earlier incorrect variant.

**README missing glance docs** — also covered by `a26f0b7`; `mode: glance` has had its own section and mode-hint entry since that commit.

**`mode: all` in production** — fair point. I've just pushed `cb8779d` which adds `mode: analyst` and `mode: all` sections to the README (same pattern as flow/glance), updates the mode-hint comment to list all five options, and removes the "dev convenience" wording from the inline comment. `mode: all` now ships as a documented, intentional mode — it stacks Glance, Flow, and Analyst in front of the classic tab set, which is genuinely useful for users who want all display styles in one dashboard.

custom_components/givenergy_local/www/ge-strategy.js
dewet22: Addressed in cb8779d: removed the 'dev convenience' label, added README sections for `mode: analyst` and `mode: all` (same pattern as flow/glance), and updated the mode-hint comment to list all five options. `mode: all` now ships as a documented, intentional mode.

PR #139, opened 2026-06-07
fix(migrate): handle HA 2026.6 area-prefixed entities in the GivTCP migration script

dewet22:
## What

I came back to `scripts/migrate_from_givtcp.py` to check it still worked before migrating my own data, and it turned out to break on HA 2026.6. Four related changes:

**Resolve area-prefixed / renamed entity IDs.** The script hard-coded its `givenergy_local` targets as `sensor.givenergy_inverter_<sn>_<suffix>`. Under HA 2026.6 the real entity IDs (and their recorder `statistic_id`s) carry the device-area prefix — `sensor.loft_givenergy_inverter_…` — so on a real install it wrote history into phantom statistics and left the actual series un-backfilled. It now rebuilds the same canonical→actual map the dashboard generator uses (`_build_entity_id_resolver`), from the WebSocket entity/device registries since the script runs out-of-process.

**Drop the battery_cycles migration.** #126 added a `BATTERY_PAIRS` entry to copy per-pack charge cycles, but GivTCP records cycles as a *mean* statistic while `givenergy_local`'s `charge_cycles` is `total_increasing` (a *sum* series). Incompatible shapes — a naive copy would rebase the GE counter to ~0 and corrupt it — so I've reverted it, with a comment recording the real reason (the original "GivTCP has no per-battery LTS" note was wrong) and a guard test.

**Refuse to write an unresolvable target.** If resolution yields an id that isn't in the recorder, `--apply` now refuses up front and names the targets, rather than writing phantom series. A real target with no pre-cutover overlap is flagged but not blocked.

**Correct the re-run safety note.** The apply warning claimed re-running with any cutover was safe; it's only idempotent with the same or a later cutover.

## Verification

Ran the full flow against my live HA — auto-detect, `--cutover` preview, and `--apply`. ~18 months of GivTCP history migrated, the seam is continuous (no sum cliff), and the Energy dashboard renders the extended history. Added regression tests for the resolver, the slugify equivalence against `homeassistant.util.slugify`, the write guard, and the battery_cycles omission; full suite green.

No `custom_components/` changes — this is all the out-of-tree helper script, so the HACS package is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Migration script now properly resolves Home Assistant entity IDs from live registries, improving accuracy when migrating from GivTCP.

* **Bug Fixes**
  * Enhanced migration safety with pre-flight checks to prevent orphan series creation.
  * Removed non-functional battery cycle migration.

* **Documentation**
  * Clarified which fields are and aren't migrated during the GivTCP migration process.


dewet22-codex:

scripts/migrate_from_givtcp.py
```diff
@@ -616,13 +750,40 @@ async def run(args: argparse.Namespace) -> int:
...
+    plan = _build_plan(
+        inv_serials, batt_serials, meta_by_id, args.include_charge_from_grid, resolve
+    )
+
+    # Pre-flight guard: every resolved givenergy_local target must be a known
+    # recorder statistic. A miss means resolution produced a phantom id (an
+    # area-prefix/rename it couldn't map, or an entity with no LTS) — applying
+    # would clear nothing and import an orphan series while the real entity stays
+    # un-migrated. Refuse to --apply rather than do that silently.
+    missing_targets = [ge_id for (_givtcp, ge_id, *_rest) in plan if ge_id not in meta_by_id]
```
dewet22-codex: This pre-flight blocks on every catalogue target before checking whether the corresponding GivTCP source has anything to migrate. That rejects valid topologies: `pv_energy_today_kwh` is always in `INVERTER_PAIRS`, while its GE target `PV Energy Today` is deliberately not created on three-phase inverters (`single_phase_only=True`), so `--apply` will refuse the entire migration even though that absent target can simply be skipped when its source is irrelevant. The per-entity path already checks `no_givtcp_data` before `ge_known`; I think the up-front guard needs the same distinction, for example by only treating a missing target as fatal when its GivTCP source is present in the recorder.

RELEASE_NOTES.md
```diff
@@ -99,3 +99,105 @@ devices:
...
+
+"Grid Export Power" has been renamed to "Grid Power" and now reports signed net flow — positive when exporting, negative when importing — matching what the underlying register actually measures. The existing entity and its history are migrated in place under the new slug, so no history is lost.
+
+**Stable entity_ids for control entities**
+
+Number, select, switch and time entities now carry their device name in DeviceInfo, keeping their entity_ids stable across restarts and bringing them in line with the sensor platform.
+
+**Charge-cycle history carried over from GivTCP**
+
+The GivTCP statistics migration now copies each battery's charge-cycle count history across too, so that long-term statistic carries over when switching.
```
dewet22-codex: This newly backfilled note contradicts the migration correction in the same PR. The accompanying change explains that the v1.1.3 pair was a silent no-op because battery serial detection only scanned sum statistics while GivTCP records cycles as a mean, so charge-cycle history was not actually carried over. I think the historical note needs to describe the attempted support and its limitation, or omit this item, rather than tell v1.1.3 users that the history migrated successfully.

scripts/migrate_from_givtcp.py
```diff
@@ -157,6 +168,65 @@
...
+
+def _slugify(text: str) -> str:
+    """ASCII slugify matching homeassistant.util.slugify for GivEnergy names.
+
+    The integration's device and sensor names are plain ASCII, so lowercasing
+    and collapsing runs of non-alphanumerics to a single underscore reproduces
+    HA's slug. A test pins this against homeassistant.util.slugify for the
+    device names and every mapped sensor name.
+    """
+    return re.sub(r"[^a-z0-9]+", "_", text.lower()).strip("_")
```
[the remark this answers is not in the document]
dewet22: Thanks. In this script `_slugify` only ever sees the integration's *original* device and entity names — the resolver rebuilds the canonical id from `device.name` + `original_name`, not user-set overrides — and GivEnergy's are all ASCII, so the transliteration path isn't reachable here. I'd rather keep it ASCII-only as documented than add handling for inputs that can't occur; the test pins it against `homeassistant.util.slugify` for exactly those names.

scripts/migrate_from_givtcp.py
dewet22: Good catch — fixed in 145ec99. I narrowed the up-front abort to *systemic* failure: it bails only when not one target resolves (the area-prefix/rename case) and leaves partial misses to the per-entity `ge_not_found` path. Gating purely on "GivTCP source present" wouldn't cover this example — `pv_energy_today_kwh` is present on three-phase while its `single_phase_only` target isn't — so "nothing resolved" is what distinguishes a real resolver failure from a topology gap. Added unit coverage for the three cases.

RELEASE_NOTES.md
dewet22: Agreed — reworded in 5dc6279 to say the v1.1.3 pair was attempted but never carried cycles over (GivTCP mean vs `total_increasing`), consistent with the revert in this PR, rather than implying it worked.

dewet22-codex:

scripts/migrate_from_givtcp.py
```diff
@@ -507,26 +614,51 @@ def _build_plan(
...
+    or the entity is gated off on this topology — e.g. `single_phase_only`
+    sensors (PV Energy Today, House Consumption Today) are never created on a
+    three-phase inverter. Those are skipped per-entity (`ge_not_found`) without
+    blocking the rest. But if NOT ONE target resolved, the resolver itself
+    failed (an area prefix / rename it couldn't map), so `--apply` would write
+    only phantom series — surface every miss for an up-front abort. Returns []
+    as soon as at least one target resolves.
+    """
+    missing = [ge_id for (_givtcp, ge_id, *_rest) in plan if ge_id not in known_ids]
+    resolved_any = any(ge_id in known_ids for (_givtcp, ge_id, *_rest) in plan)
```
dewet22-codex: This can still miss the exact systemic failure the PR is intended to repair. A user who already ran the broken area-prefix version may have one or more orphan canonical statistics in the recorder; those phantom ids are therefore in `known_ids`. If the registry resolver now fails for every real area-prefixed target, that old orphan makes `resolved_any` true, the pre-flight permits `--apply`, and the script clears/reimports the phantom again while skipping the real targets as `ge_not_found`. I think the systemic check needs evidence that the registry resolver mapped a real target, rather than treating mere presence of a planned statistic id as proof of successful resolution.

dewet22: Good catch — you're right that recorder presence isn't proof of resolution; an orphan from a prior broken `--apply` lives in the recorder too. Fixed in 35f5348: I now carry a registry-recognition flag through the plan (`canonical in canonical_to_actual`) and key both the pre-flight and the per-entity `ge_known` guard on that, never on recorder presence — so an orphan can't pass for a real target. Verified live: an empty resolver still refuses `--apply` (exit 2, no writes), and the healthy preview resolves unchanged.

dewet22-codex:

dewet22-codex (approved):
Approved. I re-reviewed the current merged head, including the registry-recognition fix for orphan statistics, and have no remaining actionable concerns.

dewet22:
I've just released **v1.1.5**, and that push reordered `RELEASE_NOTES.md` to reverse-chronological (newest first) and backfilled v1.1.0–v1.1.5 — so the notes edits on this branch now conflict and are superseded. Simplest rebase is to drop the `RELEASE_NOTES.md` changes here entirely and take main's version.

The migration fix isn't in the released notes yet, and v1.1.5 is now taken (glance + analyst modes, the flow centre-label fix, and the `generate_dashboard` deprecation). If it warrants an entry, it'd be a fresh `## v1.1.6` at the top — though as a `scripts/` helper it may just land on main without a tag.

The script/doc/test changes are the useful part here and look solid; only the notes need the rebase.

dewet22-codex:
I’ve started re-reviewing this PR after the rebase.

dewet22-codex (approved):
Approved after the rebase. I reviewed the new head, including the retained registry-recognition safeguards and the post-rebase diagnostic wording change, and have no actionable concerns.

scripts/migrate_from_givtcp.py
```diff
@@ -557,8 +706,14 @@ def _print_summary(results: list[MigrationResult], applying: bool) -> int:
     for r in results:
         if r.status == "error":
             print(f"\n  ERROR — {r.description}: {r.error}", file=sys.stderr)
+        elif r.status == "ge_not_found":
+            print(
+                f"\n  ⚠️  {r.description}: givenergy_local target {r.ge_id} not found in "
+                "recorder statistics — resolution failed or the entity has no LTS. Skipped.",
+                file=sys.stderr,
```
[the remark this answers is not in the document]
dewet22: Good spot — fixed in 3f68570. The message now matches the registry-recognition guard: the target "was not resolved from the entity registry (area prefix / rename unmapped, or no such entity)".

scripts/migrate_from_givtcp.py
```diff
@@ -616,13 +780,42 @@ async def run(args: argparse.Namespace) -> int:
...
         print("  ⚠️  listed givenergy_local entities. Back up your recorder")
         print("  ⚠️  database before proceeding.")
-        print("  ℹ️  Re-running (same or different --cutover) is safe: the script")
-        print("  ℹ️  always reads the original GivTCP entity for pre-cutover data.")
+        print("  ℹ️  Re-running with the SAME or a LATER --cutover is safe and")
+        print("  ℹ️  idempotent (the script always re-reads the original GivTCP entity")
+        print("  ℹ️  for pre-cutover data). An EARLIER cutover is not — it would fold")
+        print("  ℹ️  already-migrated history back through the rebase. If you use")
+        print("  ℹ️  today's date, the day's stats are still live, so wait until")
+        print("  ℹ️  tomorrow before re-running.")
```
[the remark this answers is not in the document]
dewet22: Thanks, but this project's ruff config selects only `E, F, I, UP` — the `RUF` ruleset (and so RUF001) isn't enabled, which is why CI is green. The `ℹ️`/`⚠️` markers are intentional, consistent style in this script (the `ℹ️` predates this change), so I'll keep them as-is.

PR #140, opened 2026-06-07
Surface a missing device instead of silently baking in the loss

dewet22:
## What

When `detect(prior=…)` finds a previously-known device missing, the coordinator no longer silently bakes in the reduced topology. It now:

- **Classifies** the mismatch — a *device loss* (a known battery/meter/HV stack absent, `device_type` unchanged) versus a routine add / `device_type` change.
- **Retries** a loss a couple of times (5s apart) before believing it — a slow BMS often answers on the next sweep.
- On a **persistent loss**, serves the reduced topology for that poll only but keeps the full prior cached (no persist, no reload), so the next reconnect re-probes the full set and self-heals. The missing device's entities read unavailable rather than vanishing.
- Raises a **loud, fixable ERROR repair** naming the missing device(s). It clears automatically once the device returns, or you click Fix to clear the cache and re-detect (committing a genuine removal).

Adds and `device_type` changes keep the existing persist-and-reload path unchanged.

## Why

This came out of a real incident: a restart triggered a detect that transiently missed a battery (its BMS was slow to answer one probe window). The integration dropped the pack, its entities went unavailable, and nothing was recorded for it until I spotted it on a dashboard — `redetect_plant` recovered it, since the pack had been responding the whole time. The worst case is a restart where none of a known device's data is recorded until a human notices.

## Scope

This is the hass-side resilience change. The dashboard Redetect button and serial-targeted services already shipped. A follow-up modbus change (a placeholder so a device missed at cold setup stays present-but-unavailable within the same session, rather than absent until the next restart) isn't needed for this to be safe — today a missing device just reads unavailable, never crashes — so I've deferred it to a separate handoff.

## Testing

`uv run pytest` (266 pass, 9 new), ruff and mypy clean. New tests cover the classifier, retry-then-heal, persistent loss (callback fires, prior kept, nothing persisted/reloaded), adds and `device_type` changes (unchanged path), a retry that surfaces an add, the ERROR repair creation, live self-heal, and the fix flow. I haven't validated against the real hardware yet — I'll do that via the HA MCP once merged (reproduce a transient miss; confirm the ERROR repair appears and clears).

## Summary by CodeRabbit

* **New Features**
  * Added detection for missing GivEnergy devices with user-friendly repair guidance to clear cached layout and rerun device discovery.
  * Integration now creates fixable repair issues when expected devices stop responding.
  * Automatically clears repair issues when device connectivity is restored.

* **Bug Fixes**
  * Improved distinction between temporary and persistent device loss to prevent unnecessary configuration reloads.

### DIFF
--- a/custom_components/givenergy_local/coordinator.py
+++ b/custom_components/givenergy_local/coordinator.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import logging
 from collections.abc import Awaitable, Callable
 from datetime import datetime, timedelta
@@ -28,6 +29,17 @@ InverterModel = SinglePhaseInverter | ThreePhaseInverter
 # the coordinator itself stays free of HA UI / config-entry concerns.
 TopologyChangedCallback = Callable[[PlantCapabilities], Awaitable[None]]
 
+# Invoked when detect() reports that a previously-known device did NOT respond
+# (a loss) and the loss persisted across retries. Receives (prior, actual) so
+# the caller can raise a loud, fixable repair WITHOUT persisting the reduced
+# topology — the full prior stays cached so the next reconnect re-probes it.
+DevicesMissingCallback = Callable[[PlantCapabilities, PlantCapabilities], Awaitable[None]]
+
+# Invoked whenever a (re)connect confirms the full expected topology (detect
+# succeeded with no mismatch, or a loss healed on retry). Lets the caller clear
+# a stale "device missing" repair the moment the device answers again.
+TopologyHealedCallback = Callable[[], Awaitable[None]]
+
 _LOGGER = logging.getLogger(__name__)
 
 # Target interval between full holding-register refreshes in active mode.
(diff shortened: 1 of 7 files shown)


dewet22-codex:

custom_components/givenergy_local/coordinator.py
```diff
@@ -378,25 +437,95 @@ async def _connect(self) -> None:
...
+            # the reduced topology for this tick only so the live poll dispatches
+            # for what responded; the missing device's entities read unavailable.
+            _LOGGER.error(
+                "Expected device(s) %s did not respond after %d retries; serving the "
+                "reduced topology for this poll but NOT persisting it — affected "
+                "entities will read unavailable until they reappear or you re-detect",
+                missing,
+                DETECT_LOSS_RETRIES,
+            )
+            self._client.plant.capabilities = exc.actual
```
dewet22-codex, line 516: This leaves the reduced client connected after the retry window. `_async_update_data()` only calls `_connect()` and `detect()` when the client is absent or disconnected, while normal refreshes now dispatch from `exc.actual`; a battery that starts answering again after these retries is therefore never probed, and the repair cannot clear until an unrelated disconnect, restart, or manual Fix. That can still leave its history unrecorded indefinitely after a transient miss, which is the incident this change is intended to prevent. I think the persistent-loss state needs to arrange a later re-detect or reconnect with suitable backoff, rather than relying on a reconnect that may never occur.

dewet22:
Thanks for the reviews — both points are addressed in b35b9b3.

**Gemini** — `missing_devices(prior=None, …)` now short-circuits to `[]` immediately (matching the type hint update to `PlantCapabilities | None`), and the classification test has a `prior=None` assertion.

**Codex** — The concern is correct: after retries are exhausted the client stays connected, so `_connect()` / `detect()` would never re-run and the auto-heal via `on_topology_healed` couldn't fire. Fixed with a `_schedule_reconnect` flag: the persistent-loss path sets it, and `_async_update_data()` checks it at the top of the next tick — resetting the client before `reconnecting = …` is evaluated, which forces a fresh `_connect()` → `detect(prior=full_prior)`. If the device is back it heals; if not, we go around again. The persistent-loss test now asserts the flag is set.

dewet22-codex:

custom_components/givenergy_local/coordinator.py
```diff
@@ -378,25 +441,96 @@ async def _connect(self) -> None:
...
+            self._schedule_reconnect = True
```
dewet22-codex, line 517: This schedules another reconnect every time the loss remains persistent: the next poll reconnects and runs the initial detect plus both delayed retries, then reaches this line and schedules the same work again. For a genuinely removed or long-offline device, the default 30-second polling cycle would therefore spend roughly 10 seconds repeatedly detecting and churn the TCP connection indefinitely until the repair is applied. I think this still needs a separate, suitably backed-off re-detection schedule so recovery remains automatic without making every normal poll pay the reconnect/retry sequence.

dewet22:
Addressed in 73072dc.

You're right — `_schedule_reconnect` gets re-set each time `_handle_topology_mismatch` reaches the persistent-loss path, so without a cooldown the coordinator would spend ~10 s on detect+retries every 30 s indefinitely for a genuinely absent device.

Fix: `LOSS_REDETECT_INTERVAL = 300.0 s` and a `_loss_redetect_after` float. The reconnect block in `_async_update_data()` now only fires when the running-loop clock is past that deadline:

```python
loop = asyncio.get_running_loop()
if self._schedule_reconnect and loop.time() >= self._loss_redetect_after:
    self._schedule_reconnect = False
    await self._reset_client()
    self._loss_redetect_after = loop.time() + LOSS_REDETECT_INTERVAL
```

- First loss: `_loss_redetect_after = 0.0`, so the reconnect fires immediately on the next tick (no delay on first detection).
- Persistent loss after that reconnect: cooldown set to `now + 300 s`; subsequent polls skip the reset and continue serving the reduced topology until the window expires.
- Device returns mid-window: heals on the next cooldown tick (at most 5 min wait).
- Fix button: calls `_capabilities_store.async_remove()` + `async_schedule_reload()` directly, so it triggers an immediate redetect without waiting for the cooldown.

New test `test_loss_reconnect_respects_cooldown` verifies that with `_loss_redetect_after = float("inf")`, a `_schedule_reconnect = True` flag does not reset the client.

dewet22-codex:

dewet22-codex (approved):
Approved. The loss-driven reconnect now retains automatic recovery while limiting repeated detection work to the five-minute cooldown, which addresses my remaining concern. I found no further actionable issues in the current head.

PR #141, opened 2026-06-07
Remove deprecated generate_dashboard service

dewet22:
## Summary

- Deletes `dashboard.py` (999 lines) and `tests/test_dashboard.py` (367 lines)
- Removes `SERVICE_GENERATE_DASHBOARD` and its voluptuous schema, service registration, and unload hook
- Removes the dashboard store (`Store`) and `dashboard_outdated` repair issue — no new outdated-schema repairs will be raised; any lingering ones on existing installs will remain dismissable
- Removes `_missing_dashboard_cards`, `_REQUIRED_HACS_CARDS`, `_build_entity_id_resolver` (only used by the service handler)
- Removes `DashboardOutdatedRepairFlow` from `repairs.py`; unknown fixable issue IDs now raise `ValueError` rather than silently falling through to the dashboard flow
- Removes `dashboard_outdated` from `strings.json` and `translations/en.json`
- Removes the two dashboard-specific tests from `tests/test_script_entity_refs.py`; the migration-script entity-ref tests remain
- Net: −1731 lines; 235 tests pass

The live `custom:givenergy` strategy card has been shipping since v1.1.4. The service was deprecated in v1.1.5 with a warning log and a deprecation notice in the notification it produced. This removes it completely.

The `redetect_plant`, `capture_frames`, and `expose_recommended_entities` services are unaffected.

## Test plan

- [ ] `uv run pytest` passes (235 tests)
- [ ] `uv run mypy custom_components` clean (0 errors)
- [ ] `uv run ruff check && uv run ruff format` clean
- [ ] HA loads without error after install (no `generate_dashboard` service in developer tools)
- [ ] Existing `dashboard_outdated` repair issues (if any) remain dismissable

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


dewet22-codex:

custom_components/givenergy_local/services.yaml
```diff
@@ -150,29 +150,3 @@ redetect_plant:
       required: false
       selector:
         text:
-
-generate_dashboard:
```
dewet22-codex, line 154: The bundled live strategy still emits a “Regenerate Dashboard” button whose `perform_action` is `givenergy_local.generate_dashboard` (`www/ge-strategy.js:1082-1088`). After this service is removed, every classic/all strategy dashboard will retain a visible control that fails with “service not found”. I think the strategy and its JS tests need to lose or replace that action as part of this removal.

custom_components/givenergy_local/dashboard.py
```diff
@@ -1,999 +0,0 @@
-"""Generate a dashboard YAML for the GivEnergy Local integration."""
```
dewet22-codex, line 1: Deleting this module leaves the checked-in `dashboard/generate.py` helper broken: it loads this exact path with `spec_from_file_location()` and then accesses `generate_dashboard`, so it will fail immediately for anyone still using the documented standalone generator. I think that helper needs to be removed in the same change, or retained with an implementation that no longer depends on this module.

custom_components/givenergy_local/services.yaml
```diff
@@ -150,29 +150,3 @@ redetect_plant:
...
-  name: Generate dashboard (deprecated)
```
dewet22-codex, line 155: The current user-facing guidance still presents this service as available: `README.md` lists it in the service table and retains the deprecated static-dashboard section, while the `ems_entity_ids_changed` text in both `strings.json` and `translations/en.json` tells users to run “Generate dashboard” after recreating IDs. Once this service is removed, that repair workflow directs users to an action they cannot perform. I think those references need updating alongside the service removal.

dewet22-codex:

custom_components/givenergy_local/repairs.py
```diff
@@ -17,30 +15,7 @@ async def async_create_fix_flow(
...
-            return self.async_create_entry(data={})
-        return self.async_show_form(
-            step_id="init",
-            data_schema=vol.Schema({}),
-            description_placeholders={
-                "old_version": str(self._data.get("old_version", "?")),
-                "new_version": str(self._data.get("new_version", "?")),
-            },
-        )
+    raise ValueError(f"Unknown fixable issue: {issue_id}")
```
dewet22-codex, line 18: This conflicts with the PR’s promise that lingering `dashboard_outdated` repairs remain safely dismissable. Existing installs can retain one of those issues with `is_fixable=True`, because setup no longer deletes the old versioned IDs; clicking its visible Fix action calls this function and the `ValueError` escapes through Home Assistant’s repairs flow manager as a server error. Returning `None` would not satisfy the current repairs protocol either, so I think the removal needs to delete the known legacy issue IDs during setup, or retain a valid flow that lets the user close them cleanly.

dewet22-codex:

custom_components/givenergy_local/__init__.py
```diff
@@ -453,29 +410,11 @@ def _migrate_unique_ids(hass: HomeAssistant, entry: ConfigEntry) -> None:
...
-            continue
-        canonical = f"{ent.domain}.{slugify(device.name)}_{slugify(ent.original_name)}"
-        canonical_to_actual[canonical] = ent.entity_id
-    return lambda entity_id: canonical_to_actual.get(entity_id, entity_id)
-
-
 async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
+    # Clear any legacy dashboard_outdated issues left by the now-removed
+    # generate_dashboard service so the Repairs UI doesn't show a broken Fix button.
+    ir.async_delete_issue(hass, DOMAIN, f"dashboard_outdated_{entry.entry_id}")
```
dewet22-codex: This does not match the IDs the removed code created. Dashboard repair issues were named `dashboard_outdated_v{DASHBOARD_VERSION}` (for example `dashboard_outdated_v11`), not `dashboard_outdated_{entry_id}`, so this deletion is a no-op and an existing issue can still expose the broken Fix action. Direct upgrades may also retain an issue from an earlier schema version. I think this cleanup needs to delete the known versioned IDs, covering every dashboard schema version that may have raised one.

dewet22-codex:

dewet22-codex (approved):
Approved. The removal now clears every versioned legacy dashboard repair, removes the broken strategy and standalone-generator paths, and updates the remaining user-facing guidance. I found no further actionable concerns in the current head.

custom_components/givenergy_local/services.yaml
dewet22: Removed in `c99453c` — the entire button block (lines 1082–1088 in ge-strategy.js) is gone, along with the vitest fixture that exercised it.

custom_components/givenergy_local/dashboard.py
dewet22: Removed in `c99453c` — `dashboard/generate.py` and `dashboard/template.yaml` deleted via `git rm`.

custom_components/givenergy_local/services.yaml
dewet22: README service table entry and deprecated static-dashboard section removed (`95a94be`); EMS repair description updated to drop the 'run the Generate dashboard' step (`56ef95e`).

custom_components/givenergy_local/repairs.py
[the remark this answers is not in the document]
dewet22: Addressed by proactively deleting all `dashboard_outdated_v1`–`v11` issues in `async_setup_entry` (`435424c`) so no existing issue reaches this function.

dewet22: Cleared on setup in `435424c` (loops `range(1, 12)` over `dashboard_outdated_v{n}`) — the only valid consumer of `async_create_fix_flow` is `expected_devices_missing_*`, anything else is a programming error.

custom_components/givenergy_local/__init__.py
dewet22: Fixed in `435424c` — loops `range(1, 12)` over `dashboard_outdated_v{n}` to cover all 11 schema versions the old code could have emitted.

PR #143, opened 2026-06-08
fix(sensor): use TOTAL state class for computed e_consumption_today

### LINKED ISSUE #142: Has state class total_increasing, but its state is not strictly increasing
> **Fixed in v1.1.7.** The warning was caused by transient multi-register polling skew: `e_consumption_today` is computed from PV generation, grid import/export, and AC charge registers polled within the same scan cycle, and when one updates fractionally ahead of the others the derived total can dip by a few Wh. An intra-day monotonic filter now prevents those dips from reaching the recorder, while still letting the genuine midnight counter reset through. See the [v1.1.7 release notes](https://github.com/dewet22/givenergy-hass/blob/v1.1.7/RELEASE_NOTES.md#v117) for full details.

```
Log details (WARNING)
Logger: homeassistant.components.sensor.recorder
Source: components/sensor/recorder.py:431
Integration: Sensor ([documentation](https://www.home-assistant.io/integrations/sensor), [issues](https://github.com/home-assistant/core/issues?q=is%3Aissue+is%3Aopen+label%3A%22integration%3A+sensor%22))
First occurred: 10:35:10 am (1 occurrence)
Last logged: 10:35:10 am

Entity sensor.solar_and_battery_storage_givenergy_inverter_cexxxxxxxx_house_consumption_today from integration givenergy_local has state class total_increasing, but its state is not strictly increasing. Triggered by state 4.6 (previous state: 4.7) with last_updated set to 2026-06-08T09:30:37.794676+00:00. Please create a bug report at https://github.com/dewet22/givenergy-hass/issues
```

dewet22:
## What

Changes `e_consumption_today` from `SensorStateClass.TOTAL_INCREASING` to `SensorStateClass.TOTAL`.

## Why

`e_consumption_today` is a computed field — givenergy-modbus derives it as `PV gen + grid-in - grid-out - AC-charge` from several registers polled at slightly different times. When one component updates before the others, the result can transiently dip by a few Wh, triggering HA's strictly-increasing guard and logging the warning reported in #142.

The other daily energy sensors (`e_pv_day`, `e_grid_in_day`, `e_grid_out_day`, the battery day totals) are direct hardware counters that only increment, so they correctly stay `TOTAL_INCREASING`. Only the computed sensor needs the relaxed class.

Both `TOTAL` and `TOTAL_INCREASING` produce sum-type long-term statistics in HA's recorder, so existing history is unaffected.

Fixes #142

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -504,11 +504,16 @@ INVERTER_SENSORS: tuple[GivEnergyInverterSensorDescription, ...] = (
         # register; givenergy-modbus computes it (PV gen + grid-in - grid-out -
         # AC-charge). Three-phase has no such field, so skip_if_none drops it
         # there (and the value_fn getattr keeps it None-safe).
+        # TOTAL (not TOTAL_INCREASING) because the value is computed from several
+        # registers polled at slightly different times; a reading can transiently
+        # dip by a few Wh when one component updates before the others, which
+        # triggers HA's strictly-increasing guard. Direct hardware counters (PV,
+        # grid, battery) stay TOTAL_INCREASING since they never decrease.
         key="e_consumption_today",
         name="House Consumption Today",
         native_unit_of_measurement=UnitOfEnergy.KILO_WATT_HOUR,
         device_class=SensorDeviceClass.ENERGY,
-        state_class=SensorStateClass.TOTAL_INCREASING,
+        state_class=SensorStateClass.TOTAL,
         value_fn=lambda inv: getattr(inv, "e_consumption_today", None),
         skip_if_none=True,
     ),


dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -504,11 +504,16 @@ class GivEnergyCoordinatorSensorDescription(SensorEntityDescription):
...
+        # registers polled at slightly different times; a reading can transiently
+        # dip by a few Wh when one component updates before the others, which
+        # triggers HA's strictly-increasing guard. Direct hardware counters (PV,
+        # grid, battery) stay TOTAL_INCREASING since they never decrease.
         key="e_consumption_today",
         name="House Consumption Today",
         native_unit_of_measurement=UnitOfEnergy.KILO_WATT_HOUR,
         device_class=SensorDeviceClass.ENERGY,
-        state_class=SensorStateClass.TOTAL_INCREASING,
+        state_class=SensorStateClass.TOTAL,
```
dewet22-codex, line 516: `TOTAL` only starts a new recorder statistics cycle when the entity supplies a changed `last_reset`; this entity does not supply one. With this change, the normal midnight drop from the day’s final consumption to zero is therefore recorded as a negative delta and subtracts the whole day from the accumulated `sum`, whereas `TOTAL_INCREASING` currently detects that drop as a reset automatically. Brief intra-day dips would cancel once the derived value catches up, but the daily reset will not. I think this needs a matching daily `last_reset` implementation (or a different approach that preserves reset detection) before switching the state class.

PR #144, opened 2026-06-08
fix(sensor): floor e_consumption_today to 1 dp to prevent TOTAL_INCREASING dips

### LINKED ISSUE #142: Has state class total_increasing, but its state is not strictly increasing

dewet22:
## What

Floors `e_consumption_today` to one decimal place in the `value_fn` and adds `suggested_display_precision=1` to match peer daily-energy sensors. Keeps (restores) `SensorStateClass.TOTAL_INCREASING`.

## Why

`e_consumption_today` is a computed field — givenergy-modbus derives it as `PV gen + grid-in - grid-out - AC-charge` from several registers polled at slightly different times. When one component updates before the others, the result can transiently dip by a few Wh (e.g. 4.7 → 4.6 kWh), tripping HA's strictly-increasing guard (#142).

Flooring to 1 dp makes sub-0.1 kWh fluctuations invisible. Keeping `TOTAL_INCREASING` is important: switching to `TOTAL` (as in the now-reverted #143) would silently corrupt the accumulated `sum` on the midnight reset to zero, because `TOTAL` requires an explicit `last_reset` attribute to distinguish a reset from a decrease.

Display precision of 1 dp matches `e_grid_in_day`, `e_grid_out_day`, and `e_pv_day`.

Fixes #142

## Release Notes

* **Bug Fixes**
  * Updated energy consumption sensor to consistently display values with 1 decimal place precision
  * Improved handling of missing data in energy consumption reporting
  * Refined sensor behavior classification for more accurate consumption tracking

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import logging
 from collections.abc import Callable
 from dataclasses import dataclass, field
+from datetime import date
 from typing import Any
 
 from givenergy_modbus.model.battery import Battery, BatteryMaintenance
@@ -35,7 +36,9 @@ from homeassistant.const import (
 from homeassistant.core import HomeAssistant
 from homeassistant.helpers.device_registry import DeviceInfo
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
+from homeassistant.helpers.restore_state import RestoreEntity
 from homeassistant.helpers.update_coordinator import CoordinatorEntity
+from homeassistant.util import dt as dt_util
 
 from .const import DOMAIN
 from .coordinator import GivEnergyUpdateCoordinator, InverterModel
@@ -52,6 +55,10 @@ class GivEnergyInverterSensorDescription(SensorEntityDescription):
     # underlying field is single-phase-only (e.g. a p_pv1+p_pv2 sum) and so would
     # be meaningless or surface as a permanently-unavailable orphan entity.
     single_phase_only: bool = False
+    # If True, the entity's native_value is clamped to never decrease within a
+    # session. Use for computed TOTAL_INCREASING sensors whose source value can
+    # transiently dip due to multi-register polling skew.
+    monotonic: bool = False
 
 
 @dataclass(frozen=True, kw_only=True)
@@ -504,16 +511,16 @@ INVERTER_SENSORS: tuple[GivEnergyInverterSensorDescription, ...] = (
         # register; givenergy-modbus computes it (PV gen + grid-in - grid-out -
         # AC-charge). Three-phase has no such field, so skip_if_none drops it
         # there (and the value_fn getattr keeps it None-safe).
-        # TOTAL (not TOTAL_INCREASING) because the value is computed from several
-        # registers polled at slightly different times; a reading can transiently
-        # dip by a few Wh when one component updates before the others, which
-        # triggers HA's strictly-increasing guard. Direct hardware counters (PV,
-        # grid, battery) stay TOTAL_INCREASING since they never decrease.
+        # monotonic=True because the value is computed from several registers
+        # polled at slightly different times; a reading can transiently dip by
+        # a few Wh when one component updates before the others, tripping
+        # TOTAL_INCREASING's strictly-increasing guard (#142).
         key="e_consumption_today",
         name="House Consumption Today",
         native_unit_of_measurement=UnitOfEnergy.KILO_WATT_HOUR,
         device_class=SensorDeviceClass.ENERGY,
-        state_class=SensorStateClass.TOTAL,
+        state_class=SensorStateClass.TOTAL_INCREASING,
+        monotonic=True,
         value_fn=lambda inv: getattr(inv, "e_consumption_today", None),
         skip_if_none=True,
     ),
@@ -1112,7 +1119,19 @@ def _derive_display_precision(description: SensorEntityDescription, model: Any) 
     return model.precision_of(description.key)
 
 
-class GivEnergyInverterSensor(CoordinatorEntity[GivEnergyUpdateCoordinator], SensorEntity):
+# A drop larger than this (in the sensor's native unit) is treated as a
+# genuine source reset rather than a transient polling artefact. Covers the
+# case where the inverter clock lags HA's local midnight by one scan interval,
+# so the actual counter reset arrives on a subsequent read after the day
+# boundary has already been committed. The value is intentionally conservative:
+# transient multi-register skew is a few Wh; a 500 Wh floor is far above any
+# realistic polling noise while still detecting any meaningful midnight reset.
+_MONOTONIC_RESET_THRESHOLD = 0.5  # kWh (matches e_consumption_today's native unit)
+
+
+class GivEnergyInverterSensor(
+    CoordinatorEntity[GivEnergyUpdateCoordinator], SensorEntity, RestoreEntity
+):
     _attr_has_entity_name = True
     entity_description: GivEnergyInverterSensorDescription
 
@@ -1123,6 +1142,8 @@ class GivEnergyInverterSensor(CoordinatorEntity[GivEnergyUpdateCoordinator], Sen
     ) -> None:
         super().__init__(coordinator)
         self.entity_description = description
+        self._monotonic_max: float | None = None
+        self._monotonic_date: date | None = None
         precision = _derive_display_precision(description, coordinator.data.inverter)
         if precision is not None:
             self._attr_suggested_display_precision = precision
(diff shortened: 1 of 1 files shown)


dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -504,17 +505,21 @@ class GivEnergyCoordinatorSensorDescription(SensorEntityDescription):
...
         native_unit_of_measurement=UnitOfEnergy.KILO_WATT_HOUR,
         device_class=SensorDeviceClass.ENERGY,
-        state_class=SensorStateClass.TOTAL,
-        value_fn=lambda inv: getattr(inv, "e_consumption_today", None),
+        state_class=SensorStateClass.TOTAL_INCREASING,
+        suggested_display_precision=1,
+        value_fn=lambda inv: (
+            None
+            if (v := getattr(inv, "e_consumption_today", None)) is None
+            else math.floor(v * 10) / 10
```
dewet22-codex: This does not prevent the warning reported in #142: the recorded transition is already `4.7 → 4.6`, and flooring both values to one decimal still emits `4.7 → 4.6`. More generally, even a few-Wh dip crosses a bucket whenever the prior value is near a tenth boundary (for example `4.70 → 4.69` becomes `4.7 → 4.6`). Restoring `TOTAL_INCREASING` avoids PR 143’s midnight-reset regression, but I think the fluctuation handling needs a stateful monotonic filter or another approach that cannot emit intra-day decreases; this quantisation alone does not fix the original bug.

dewet22: Good spot — you're right that flooring to 1 dp cannot prevent a 4.7 -> 4.6 transition. I've reworked this as a stateful monotonic filter instead (pushed as an amended commit).

The entity now tracks a `_monotonic_max` across its lifetime and returns `max(new_reading, session_max)` from `native_value`. A new `monotonic: bool` flag on `GivEnergyInverterSensorDescription` gates the behaviour — only `e_consumption_today` sets it, so there's no risk of accidentally clamping sensors that legitimately decrease.

`TOTAL_INCREASING` is kept (as you noted in the earlier comment on #143 — switching to `TOTAL` without `last_reset` would corrupt the midnight reset to zero).

dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -504,17 +505,21 @@ class GivEnergyCoordinatorSensorDescription(SensorEntityDescription):
...
+        ),
```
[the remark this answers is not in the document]
dewet22: The floor() approach has been dropped entirely — the amended commit switches to a stateful `max()` filter instead. `_monotonic_max` is tracked on the entity instance and `native_value` returns `max(new_reading, session_max)`, so floating-point representation of the floor operand is no longer a concern.

custom_components/givenergy_local/sensor.py
```diff
@@ -1140,7 +1145,11 @@ def __init__(
 
     @property
     def native_value(self) -> Any:
-        return self.entity_description.value_fn(self.coordinator.data.inverter)
+        value = self.entity_description.value_fn(self.coordinator.data.inverter)
+        if self.entity_description.monotonic and isinstance(value, (int, float)):
+            self._monotonic_max = max(value, self._monotonic_max or value)
```
dewet22-codex: This clamp also hides the legitimate midnight reset that `TOTAL_INCREASING` needs to observe. The entity normally remains alive across midnight, so when the source drops from yesterday’s final total to zero, `_monotonic_max` still contains yesterday’s value and this returns that value instead. The sensor and recorder then stay pinned at yesterday’s total until today’s consumption exceeds it, potentially for the entire day, and no new recorder cycle is detected. I think the filter needs an explicit day boundary reset before applying the intra-day maximum, so midnight is emitted as a real decrease while smaller same-day dips remain hidden.

dewet22: Fixed in the amended commit. The filter now tracks `_monotonic_date` alongside `_monotonic_max`. On each call to `native_value`, we compare against `dt_util.now().date()` (HA's configured timezone). When the day changes, both fields reset to the incoming value — so the midnight drop to zero passes through as a real decrease and TOTAL_INCREASING starts a new recorder cycle. Intra-day dips are still hidden by the `max()` clamp as before.

dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -1123,6 +1129,8 @@ def __init__(
     ) -> None:
         super().__init__(coordinator)
         self.entity_description = description
+        self._monotonic_max: float | None = None
```
dewet22-codex, line 1145: This maximum is lost whenever the entity is recreated, including an integration reload or Home Assistant restart during the day. If the first reading after that reload is one of the transiently lower values this change is intended to hide, it becomes the new `_monotonic_max`, while recorder still has the higher pre-reload state; the original `TOTAL_INCREASING` dip warning can therefore recur, and the displayed value stays lower until it catches up. I think the filter needs to restore or seed the same-day maximum from the entity’s prior state (while still discarding it when that state belongs to the previous local day), rather than always starting from `None`.

dewet22: Fixed. `GivEnergyInverterSensor` now inherits `RestoreEntity` and overrides `async_added_to_hass`: for entities with `monotonic=True` it calls `async_get_last_state()` and, if the last persisted state is from today (checked via `dt_util.as_local(last_state.last_updated).date()`), seeds `_monotonic_max` and `_monotonic_date` from it. A restart mid-day therefore picks up from the last recorded maximum, so a transient-dip first reading can't undercut the recorder's last value. States from a previous day are ignored, letting the midnight reset through as normal.

dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -1138,9 +1149,40 @@ def __init__(
...
+            restored = float(last_state.state)
+        except ValueError, TypeError:
+            return
+        last_date = dt_util.as_local(last_state.last_updated).date()
+        if last_date == dt_util.now().date():
+            # Seed the intra-day max from the last persisted value so a
+            # transient dip on the first post-restart reading doesn't become
+            # the new baseline while the recorder still holds the higher value.
+            self._monotonic_max = restored
+            self._monotonic_date = last_date
```
dewet22-codex, line 1179: This still assumes the inverter counter has reset before the first poll after Home Assistant's local midnight. The integration explicitly supports correcting a drifted inverter clock, so if the inverter is even one scan interval behind, the first new-day reading can still be yesterday's high value; that seeds the monotonic maximum for the new HA date, and the real reset shortly afterwards is then clamped for the whole day. I think the reset detection needs to account for the source reset independently of the HA date (for example, by allowing a sufficiently large drop around the day boundary), otherwise the midnight failure remains for drifted clocks.

dewet22: Fixed. Added a `_MONOTONIC_RESET_THRESHOLD = 0.5` (kWh) constant and an `elif` branch in `native_value`: when on the same HA date but `value < _monotonic_max - threshold`, the drop is treated as a genuine source reset and `_monotonic_max` is updated to the new value. This fires independently of the date check — so a drifted inverter clock that seeds an inflated day-one max still gets corrected when the actual counter reset arrives on the next poll. The 0.5 kWh threshold is far above realistic polling skew (a few Wh) while well below any meaningful daily consumption total.

dewet22-codex:

dewet22-codex (approved):
The delayed inverter-clock reset concern is addressed by the large-drop guard, and I have no further concerns with this revision.

PR #146, opened 2026-06-09
fix(capture): open notification link in a new tab so the SPA router can't hijack it

dewet22:
## What this does

Fixes the capture-complete notification link, which left-clicked through to the dashboard instead of opening the capture, and documents the landing page now that it works end-to-end.

### The bug

The notification linked to the landing page with a **markdown link**. Home Assistant's frontend intercepts same-origin markdown-link clicks and routes them through the SPA router; `/api/givenergy_local/capture/…` matches no panel, so the click landed on `/home/dashboard` (carrying `?authSig=` along). Middle-click worked because it bypasses the router entirely.

### The fix

Emit a raw `<a target="_blank">` instead of a markdown link. Tracing the HA frontend source, a truthy `target` is the one anchor attribute that both:

- survives the persistent-notification markdown sanitiser (js-xss keeps `target`/`href`/`title`; strips `rel` and `download`), and
- makes `isNavigationClick` skip the click so the browser performs a real navigation to the backend view.

The signed URL stays path-only, so the new tab resolves `/api/…` against the user's actual origin — correct for reverse-proxied instances (the earlier `get_url` approach broke those with an insecure-connection warning).

### Also here

- **Docs** — screenshot of the landing page and a reframed wire-capture section (inspect / download / file-issue, not just a bare download link).
- **Dependency floor** — bump `givenergy-modbus` to `>=2.1.6`, which removes the unreliable `first_battery_serial_number` field ([modbus#191](https://github.com/dewet22/givenergy-modbus/issues/191)). That field was the source of the ghost `HC…` battery device on AIO systems (#95).

## Testing

- New `test_notification_url_is_origin_relative` guards the path-only contract.
- The notification test now asserts `target="_blank"` and rejects a markdown link, so the SPA-hijack regression can't sneak back in.
- Full suite: 236 passing. Verified live on a reverse-proxied instance — the link now opens the capture in a new tab as intended.


dewet22-codex:

custom_components/givenergy_local/manifest.json
```diff
@@ -13,7 +13,7 @@
   "iot_class": "local_polling",
   "issue_tracker": "https://github.com/dewet22/givenergy-hass/issues",
   "requirements": [
-    "givenergy-modbus>=2.1.3,<3.0.0"
+    "givenergy-modbus>=2.1.6,<3.0.0"
```
dewet22-codex, line 16: This raises the runtime manifest floor to 2.1.6, but pyproject.toml and the project entry in uv.lock still declare >=2.1.3. That leaves local development and lock-based installs able to resolve the older library whose removed battery field this commit is intended to exclude, and breaks the repository's two-file dependency-pin invariant. I think the project constraint and regenerated lockfile need to move with the manifest here.

dewet22: Good catch — you're right that I only moved the runtime pin and left the dev/lock side behind. Fixed in 73582fa: bumped the `pyproject.toml` constraint to `>=2.1.6` and regenerated `uv.lock` (now resolving 2.1.6), so all three pins agree again. Full suite passes against 2.1.6, which also confirms the integration doesn't touch the removed field.

PR #147, opened 2026-06-09
feat(sensor): expose per-module AIO battery devices (#192)

dewet22:
## What this does

Surfaces each removable **All-in-One battery module** as its own Home Assistant device — closing the gap where AIO data was only available at pack level (the original ask in #95, now tracked as #192).

On AIO hardware, each module (device addresses `0x50–0x53`) becomes a device parented to the AIO inverter, keyed by its `HX…` serial, exposing:

- **24 per-cell voltages** (`v_cell_01…24`)
- **12 per-cell temperatures** (`t_cell_01…12` — cells 13–24 read zero on known hardware, so they're omitted)

These mirror the existing LV per-cell entities exactly: `DIAGNOSTIC`, enabled by default, so the cell-heatmap card and the same pack-health views work at module granularity.

## How it's gated

- Driven off `PlantCapabilities.aio_battery_module_addresses` via `Plant.aio_battery_modules`, so **non-AIO plants create nothing**.
- A module with a blank/invalid serial can't anchor a device, so it's skipped (the index still aligns for the modules we do create).

## Dependency

Requires the per-module API from **givenergy-modbus 2.2** (`AioBatteryModule`, `Plant.aio_battery_modules`, `DeviceType.BATTERY_MODULE`). The floor is pinned to `>=2.2.0rc3` across manifest, pyproject, and the lockfile — only givenergy-modbus resolves to a pre-release. This is intended for a **feature RC** to validate against real AIO hardware (AberDino's 4-module unit, #95) before 2.2 final.

## Testing

- 6 new tests: per-module devices created and parented to the inverter, all 24 voltage + 12 temp sensors present, temps 13–24 absent, invalid-serial module skipped, non-AIO plants create no module entities, description coverage.
- Full suite: 242 passing against 2.2.0rc3. ruff/format/mypy clean.

Not yet validated on real AIO hardware — that's the point of the RC.

## Summary by CodeRabbit

* **New Features**
  * Added support for monitoring individual battery modules in All-in-One systems with per-cell voltage and temperature diagnostics.

* **Documentation**
  * Updated to reflect v2.2 compatibility and clarified battery module representation on All-in-One hardware.

* **Dependencies**
  * Updated minimum required version to givenergy-modbus v2.2.0rc3.

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -6,6 +6,7 @@ from dataclasses import dataclass, field
 from datetime import date
 from typing import Any
 
+from givenergy_modbus.model.aio_battery import AioBatteryModule
 from givenergy_modbus.model.battery import Battery, BatteryMaintenance
 from givenergy_modbus.model.inverter import (
     BatteryCalibrationStage,
@@ -74,6 +75,20 @@ def _battery_attr(name: str) -> Callable[[Battery], Any]:
     of the resulting Callable).
     """
     return lambda bat: getattr(bat, name)
+
+
+@dataclass(frozen=True, kw_only=True)
+class GivEnergyAioModuleSensorDescription(SensorEntityDescription):
+    value_fn: Callable[[AioBatteryModule], Any] = field(default=lambda _: None)
+
+
+def _module_attr(name: str) -> Callable[[AioBatteryModule], Any]:
+    """Return a value_fn that reads `name` off an AIO battery module.
+
+    Per-module counterpart of `_battery_attr`; each closure captures its own
+    attribute name so the bulk-defined per-cell entities don't share one.
+    """
+    return lambda module: getattr(module, name)
 
 
 def _battery_hex(name: str, width: int) -> Callable[[Battery], Any]:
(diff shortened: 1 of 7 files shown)


dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -1241,6 +1301,45 @@ def native_value(self) -> Any:
...
+            serial_number=serial,
+            via_device=(DOMAIN, coordinator.data.inverter_serial_number),
+        )
+
+    @property
+    def native_value(self) -> Any:
+        modules = self.coordinator.data.aio_battery_modules
+        if self._module_index >= len(modules):
+            return None
+        return self.entity_description.value_fn(modules[self._module_index])
```
dewet22-codex: Plant.aio_battery_modules rebuilds this list from only the module-address caches that are currently present and decode successfully, so its indices are not stable across refreshes. If the first module temporarily drops out, the module at index 1 shifts to index 0 and these entities publish that module's cell values under the first module's serial/device until it returns. I think each entity needs to retain a stable identity such as module_address (or serial) and resolve the matching module on each update; a missing match can then return unavailable without cross-wiring module data.

dewet22-codex:

custom_components/givenergy_local/sensor.py
dewet22: Good catch — fixed in cb10af1. Entities now bind to the module serial and resolve the matching module on each read (reporting unavailable if it's absent), so a reindex when a module drops out can't cross-wire one module's cell data onto another's device. Added a regression test for the drop-out/reindex case.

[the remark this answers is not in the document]
dewet22: Addressed in cb10af1 — the new _module() resolver guards coordinator.data being None (returns None), so native_value no longer dereferences it directly and can't raise AttributeError.

custom_components/givenergy_local/sensor.py
```diff
@@ -1065,6 +1114,17 @@ async def async_setup_entry(
             for description in BATTERY_SENSORS
         )
 
+    # AIO per-module battery devices (#192) — empty on non-AIO plants. A module
+    # with a blank/invalid serial can't anchor a device, so skip it; the index
+    # still aligns with `aio_battery_modules` for the entities we do create.
+    for module_index, module in enumerate(coordinator.data.aio_battery_modules):
```
dewet22-codex, line 1123: This still creates entities only for modules that decoded in the first coordinator snapshot. On a warm start, a known AIO module can be persistently missing during the initial detect/retry, so the coordinator serves the reduced topology and setup reaches this loop without that module; when it answers on the scheduled reconnect, the topology-healed callback only clears the repair and does not reload the platforms. Its module device and entities therefore remain absent until a manual/integration reload, rather than appearing unavailable and recovering. I think the healed path needs to trigger entity creation/reload for this case, or setup needs another way to represent expected modules that were absent initially.

dewet22-codex:

custom_components/givenergy_local/sensor.py
dewet22: Valid point, thanks. This is the same frozen-at-setup limitation the LV battery entities already have — `missing_devices()` doesn't track `aio_battery_module_addresses` and `_on_topology_healed` clears the repair without reloading, so a device absent during the initial probe stays without entities until a reload regardless of type. Rather than bolt a partial fix onto AIO alone, I've filed #148 to fix it uniformly for batteries, meters, HV stacks, and modules (reload-on-heal when the confirmed topology exceeds what was instantiated, plus tracking modules in missing_devices). Deferring it from this PR so the AIO entities ship consistent with the existing battery behaviour; added a code comment (01946ab) pointing at #148.

PR #150, opened 2026-06-10
fix(capture): tighten capture access to admins + signed links; write files 0600

### LINKED ISSUE #149: Security review 2026-06-10 — follow-up hardening
I ran a full security review of the codebase at 488b26c — the write-up is at [docs/security-review-2026-06-10.md](https://github.com/dewet22/givenergy-hass/blob/main/docs/security-review-2026-06-10.md). Headline: no exploitable vulnerabilities. The HTTP capture views, frontend module, service handlers, config flow, migration script, and GitHub Actions were all covered.

Two low-severity, defence-in-depth observations came out of it, tracked here so the hardening PRs have an anchor:

- [ ] **Tighten capture-view access.** `requires_auth = True` admits any authenticated HA user, and capture filenames are epoch-based (guessable). Captures are already redacted so impact is minimal, but access can be tightened to admins plus the signed notification links.
- [ ] **Write capture files 0600.** They currently land with default permissions; same trust boundary as the rest of the HA config dir, but no reason to be broader than needed.

Neither is exploitable as-is — both match Home Assistant's standard trust model — so these are tightenings rather than fixes.

dewet22:
## What this does

Implements both defence-in-depth observations from the [2026-06-10 security review](https://github.com/dewet22/givenergy-hass/blob/main/docs/security-review-2026-06-10.md). Closes #149.

### Capture-view access: admin or signed link

`requires_auth = True` admits any authenticated HA user, and capture filenames are epoch-based (guessable within a narrow window), so a non-admin user could enumerate and download captures. The views now additionally require the caller to be an **admin**, or to have arrived via one of the integration's **signed links**.

The design detail worth reviewing: the notification link is signed inside a service handler, where `async_sign_path`'s fallback resolves to HA's **content user** — a read-only system user, so a naive `is_admin` gate would break the main flow. Instead, all capture links are now signed explicitly as the content user (`use_content_user=True`), and the views compare the request's resolved refresh-token id against that token. Because the auth middleware resolves a bearer-authenticated request to the *bearer's* token id (even when a junk `authSig` is appended), query-param sniffing is avoided and that bypass is closed — there's a regression test for it.

### Capture files written 0600

Files now land `0600` and the capture directory is created `0700` (newly-created directories only; existing installs keep their mode). Contents were already redacted, but there's no reason to be broader than needed on a multi-user host.

## Testing

Four new tests: non-admin bearer → 403 on both views; non-admin bearer + junk `authSig` → 403 (bypass regression); the notification link end-to-end with **no** auth → 200 (the flow that must not break); file/dir modes asserted after a capture. Existing admin and signed-URL tests continue to pass unchanged. Full suite: 247 passing; ruff/mypy clean.

The admin path and notification-link path are covered by tests; I haven't yet re-verified the notification link on a live proxied instance — happy to do that before this merges if there's any doubt.


custom_components/givenergy_local/http.py
```diff
@@ -58,6 +69,46 @@ def capture_dir(hass: HomeAssistant) -> Path:
...
+    be able to enumerate epoch-named captures. Signed links resolve to the
+    content user's refresh token — comparing the request's resolved token id
+    against it (rather than sniffing for an `authSig` query param) means a
+    junk signature appended to a bearer request still authenticates as the
+    bearer and is rejected here.
+    """
+    user = request.get(KEY_HASS_USER)
+    if user is not None and user.is_admin:
+        return True
+    return request.get(KEY_HASS_REFRESH_TOKEN_ID) == hass.data.get(STORAGE_KEY)
```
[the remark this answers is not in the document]
dewet22: Taken in fabe76c with a regression test. For the record I checked reachability against the HA source first: hass.data["http.auth"] is set unconditionally during http-component setup (not lazily on first signing), and the only auth path that leaves the token id unset — the Supervisor unix socket — runs as an admin user, so None == None couldn't fire today. The guard is still the right call so any future token-less auth path fails closed.

dewet22-codex:

custom_components/givenergy_local/http.py
```diff
@@ -120,15 +179,15 @@ async def get(self, request: web.Request, filename: str) -> web.StreamResponse:
 
         options = []
         for name in captures:
-            signed = async_sign_path(hass, landing_path(name), _SIGNED_URL_TTL)
+            signed = _sign(hass, landing_path(name))
```
dewet22-codex: A signed URL is path-bound, so arriving via the notification proves access only to the selected filename; however this branch then lists every capture and mints content-user-signed landing URLs for all of them. Anyone who obtains one notification link can therefore enumerate and open the complete capture history, despite the new gate's stated aim of preventing non-admin enumeration. I think the capture switcher and all-capture signing needs to be limited to admin requests, while a content-user signed request only receives the selected capture and its own download link.

dewet22: Good catch — the switcher did escalate a single-capture capability into the full history, and notifications are visible to non-admins, so this mattered beyond leaked links. Fixed in dbb2910 exactly as you suggest: signed-link viewers now get only the granted capture and its own download link; the full switcher is reserved for admin bearer requests. Regression test asserts a second capture's filename never appears on a signed-link page but still does for an admin.

dewet22-codex:
I’ve started the final review of this PR.

custom_components/givenergy_local/http.py
```diff
@@ -58,6 +69,52 @@ def capture_dir(hass: HomeAssistant) -> Path:
...
+
+
+def write_capture(path: Path, content: str) -> None:
+    """Write a capture file readable by the HA user only (0600).
+
+    Capture contents are already redacted, but there's no reason to leave them
+    broader than needed on a multi-user host (security review 2026-06-10).
+    Runs in an executor — blocking I/O.
+    """
+    fd = os.open(path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
```
dewet22-codex, line 113: The 0600 mode argument only applies when O_CREAT actually creates the file; if this path already exists, O_TRUNC preserves its current mode. Because filenames have one-second resolution, a same-second capture collision (or any pre-existing same-name file) can therefore leave the newly written capture readable at an older/broader mode, despite this helper's contract. I think the opened descriptor needs an explicit fchmod to 0600, with a regression test that overwrites a pre-created 0644 path.

dewet22: Quite right — the mode argument is creation-only, so a same-second collision would have kept the broader mode through O_TRUNC. Fixed in 6ddbfbb with an fchmod on the open descriptor (inside the fdopen context so the fd can't leak if it raises), and the regression test overwrites a pre-created 0644 file as you suggested.

PR #151, opened 2026-06-10
feat(sensor): split grid power into import/export for the Energy Dashboard

dewet22:
## What this does

Makes the inverter's grid power usable in HA's Energy Dashboard without the sign-inversion that loses long-term statistics.

### The problem

Our `Grid Power` sensor is `p_grid_out` — a single signed value, **positive = export**. HA's Energy Dashboard real-time grid power expects the opposite sign, so users pick its **"Inverted"** option, which spins up a derived helper entity that only accrues statistics from creation — the existing history doesn't carry over.

### The fix

Add two always-positive, single-direction power sensors — **Grid Power Import** and **Grid Power Export** — so the Dashboard's **"Two sensors"** grid option works directly: no inversion helper, no lost history. This mirrors how grid *energy* is already split into separate import/export totals.

The bidirectional `Grid Power` keeps its signed shape (the bundled flow card keys off it) but is now **hidden by default** in favour of the split sensors. Existing installs are unaffected — `entity_registry_visible_default` only applies at first registration, so anyone who already has it keeps it visible with full history.

### Naming note

They're "Grid Power Import/Export", not the more natural "Grid Import/Export Power", on purpose: `grid_export_power` is the *legacy* entity_id slug of today's `grid_power` and is actively reclaimed by the unique_id migration ([`_RENAMED_UNIQUE_ID_SUFFIXES`](custom_components/givenergy_local/__init__.py)). Reusing it would collide and silently swap semantics for a version-skipping user — there's an existing migration test guarding that slug.

## Testing

- Helper unit test covering both directions + the pre-poll None case.
- Integration test: both split sensors created with correct values (mock is importing 800 W → import 800, export 0).
- `grid_power` asserted hidden-by-default; the existing migration test still passes (no slug collision).
- Full suite: 253 passing; ruff/format/mypy clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,8 @@ When enabled, the integration connects to the inverter but sends no Modbus read 
 | Battery Charge Today | kWh | |
 | Battery Discharge Today | kWh | |
 | Battery Throughput Total | kWh | |
-| Grid Power | W | Positive = exporting, negative = importing |
+| Grid Power | W | Signed net flow: positive = exporting, negative = importing. Hidden by default (feeds the bundled flow card); for the Energy Dashboard use the split sensors below |
+| Grid Power Import / Export | W | Always-positive single-direction power, for the Energy Dashboard's "Two sensors" grid option — no sign inversion, so long-term statistics aren't lost |
 | Grid Export / Import Today | kWh | |
 | Grid Export / Import Total | kWh | |
 | AC Voltage / Frequency | V / Hz | |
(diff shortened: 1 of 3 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the split grid-power sensors and found no actionable concerns. The direction handling, registry behaviour, and legacy entity-id collision avoidance all look sound.

Issue #152, opened 2026-06-10
Mark sensors unavailable when their register block goes persistently stale


dewet22:
## Problem

The coordinator's freshness check (`_check_cache_freshness`) is **whole-plant** — it raises `UpdateFailed` only when `inverter.system_time` stops advancing. A *single* register bank that stops committing — a Pattern B all-zero rejection ([modbus#206](https://github.com/dewet22/givenergy-modbus/issues/206)), or a bank that simply stops responding while the rest of the plant polls fine — keeps serving last-good values **indefinitely**, and the sensors backed by it stay `available`. Stale-but-plausible readings silently masking a fault are worse than going unavailable; past a short staleness ceiling those sensors should drop out.

## Real sighting (#95)

AberDino's All-in-One logged 48× `Rejected all-zero IR bank (base 240) for device 0x11 over non-zero cache … keeping last-good` over a ~25-minute window before self-resolving. 25 minutes of a silently-frozen sensor is already far too long to tolerate.

**Evidence this was real data, not an empty-page ACK:**
- IR(240) is a genuine data bank — confirmed non-zero in the modbus Hybrid Gen-1 captures (served there at `0x31`; it carries `p_combined_generation`, IR 247+248, surfaced here as **Combined Generation Power**).
- The #206 guard only rejects an all-zero bank *over existing non-zero cache* — so AberDino's `0x11` IR(240) had committed real values before going all-zero. A genuine dropout of live data.
- Blast radius of that specific bank is modest (one power sensor), but it's luck-of-the-draw which bank Pattern B hits — the freshness model shouldn't depend on the dropped bank happening to be a quiet one.

## Available primitive

`givenergy-modbus` exposes `Plant.block_age(device_address, reg_type, base_register, register_count)` → seconds since that block last **committed** (`None` if never). A rejected/absent bank's age grows while the rest of the plant stays fresh — the right signal, and cleaner than content-staleness (which healthy dongle fan-out trips legitimately).

## Proposal

- When a bank a sensor depends on has a `block_age` past a ceiling, mark those sensors **unavailable**.
- **Ceiling: ~5 minutes**, expressed as `max(5 min, a small multiple of scan_interval)` so longer-interval installs aren't flagged on poll-cadence jitter. At the 30 s default that's ~10 missed commits — unambiguously "this bank stopped".
[...]

PR #158, opened 2026-06-11
feat(sensor): mark inverter sensors unavailable when their IR bank stops committing

dewet22:
Refs #152 (first cut — inverter sensors; battery/AIO-module sensors to follow).

The coordinator's freshness check is whole-plant (`system_time`), so a single register bank that stops committing — a Pattern B all-zero rejection ([modbus#206](https://github.com/dewet22/givenergy-modbus/issues/206)) or a bank that stops answering while the rest of the plant polls fine — kept serving last-good values indefinitely with its sensors still `available`. AberDino's AIO froze one bank for ~25 minutes that way (#95).

**How it works**
- At entity creation, each inverter sensor resolves its backing input-register indexes through the concrete model's register LUT (`_source_ir_registers`) — fully generic, no hardcoded bank list; what exists is decided by the model/topology, per the issue's design note.
- `ir_register_age()` (coordinator) finds the freshest *stamped* IR window covering a register by scanning the plant's commit stamps, so it tracks whatever window layout was actually polled rather than assuming 60-wide banks.
- The sensor's `available` property drops to unavailable when any backing register's age passes `max(5 min, 10 × scan_interval)` — at the 30 s default that's ten missed commits.
- Deliberate non-signals: HR-backed fields (HR banks are only re-read every few ticks by design), computed fields (no single source bank), and never-committed banks (the Pattern B shape is "committed real data, then stopped"). All keep the default coordinator availability, as does everything when the whole-plant check fails — that still wins.

**Scope**
Inverter-device sensors only for now — that's where the field sighting occurred and the blast radius is. Extending the same mechanism to battery and AIO-module sensors needs the per-device address plumbing and is a clean follow-up; I'll leave #152 open for that. The `Register._idx` access is a private-attr reach into givenergy-modbus — something to give a public accessor in the library when I'm next in there.

**Verification**
Four new tests (LUT resolution incl. three-phase, stamped-window scanning, stale→unavailable flip with ceiling, non-IR sensors untouched); full suite green (257 passed), ruff clean.


dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -1251,6 +1281,14 @@ def __init__(
         self.entity_description = description
         self._monotonic_max: float | None = None
         self._monotonic_date: date | None = None
```
dewet22-codex, line 1315: I think resolving solely from description.key leaves renamed direct-register sensors outside this protection. For example, grid_power, grid_power_import, and grid_power_export all read p_grid_out, but none of those three entity keys exists in the model LUT, so all resolve to () and continue serving stale values if p_grid_out's IR bank stops committing. A separate source-field key (or explicit source-register override) on the description would let these direct aliases participate while computed fields remain deliberately untracked.

custom_components/givenergy_local/sensor.py
```diff
@@ -1285,6 +1323,28 @@ async def async_added_to_hass(self) -> None:
...
+        IR source (computed fields, HR-backed config) keep the default
+        coordinator availability, as does a never-committed bank.
+        """
+        if not super().available:
+            return False
+        if not self._source_ir_registers:
+            return True
+        capabilities = self.coordinator.data.capabilities
+        if capabilities is None:
+            return True
```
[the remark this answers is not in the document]
dewet22: coordinator.data can't be None for a live entity here: async_config_entry_first_refresh() guarantees data before the platforms are set up, and the coordinator retains last data on failed updates. Every property on this platform (native_value directly below included) dereferences it on that basis, so I've kept available consistent rather than special-casing it.

custom_components/givenergy_local/sensor.py
dewet22: Good catch — fixed in d559c49: descriptions gain an optional source_field naming the model field the value_fn actually reads, and resolution prefers it over the key. The three grid_power* sensors map to p_grid_out and work_time_total to work_time_total_hours, with a drift test pinning the expected set. The battery charge/discharge day sensors read the canonical per-model alias (which isn't itself in the LUT), so they stay deliberately untracked alongside the other derived fields.

dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -60,6 +60,12 @@ class GivEnergyInverterSensorDescription(SensorEntityDescription):
     # session. Use for computed TOTAL_INCREASING sensors whose source value can
     # transiently dip due to multi-register polling skew.
     monotonic: bool = False
+    # Model field the value_fn actually reads, when it differs from `key` (the
+    # key is pinned by unique_id stability). Without this, renamed direct-register
+    # sensors resolve no IR source and silently miss the stale-bank availability
+    # check (#152, flagged on the #158 review). Leave None for true computed /
+    # derived fields, which are deliberately untracked.
+    source_field: str | None = None
```
dewet22-codex, line 68: The direct-alias cases are covered now, but PR #156 has since merged a topology-dependent source that this single string cannot represent: e_consumption_today is derived on SinglePhaseInverter, while on ThreePhaseInverter its value_fn falls back to the native e_load_today register. With this branch merged onto current main, the three-phase entity still resolves e_consumption_today to () and remains available when IR 1396/1397 stops committing. I think the source declaration needs to support per-model resolution (or the resolver needs a targeted e_consumption_today -> e_load_today fallback when that field exists) so the new native three-phase sensor participates without incorrectly treating the single-phase derivation as one register.

dewet22-codex:

dewet22-codex (approved):
I’ve re-reviewed the latest changes and have no concerns to add. The source-field override now resolves the three-phase native consumption registers while leaving the single-phase derived path deliberately untracked.

custom_components/givenergy_local/sensor.py
dewet22: Fixed in d5fd6d2 (after merging main in): the consumption descriptor declares source_field="e_load_today", which resolves per-model by construction — the field isn't in the single-phase LUT, so the derived value stays untracked there, while on three-phase it resolves to IR 1396–1397, exactly the register the value_fn falls back to. The drift test now pins both behaviours.

PR #161, opened 2026-06-11
fix(ci): stop the release blurb extraction deleting paragraph breaks

dewet22:
The `sed '/^$/d'` in release.yml's tag-annotation pipeline deleted every blank line from the blurb, flattening multi-paragraph release notes into one unbroken blob in the HACS update dialog — spotted on v1.3.0 (that release body has been repaired in place with `gh release edit`).

Interior blank lines are the markdown paragraph breaks between the blurb's heading+paragraph blocks, so the filter now keeps them. The PGP-signature strip and trailing-whitespace trim stay; command substitution already trims trailing newlines, which is all the non-empty annotation check needs.

Verified locally against the live v1.3.0 tag: the three interior blank lines survive the new pipeline and the emptiness check behaves as before. Real-world verification will be the next release rendering correctly in HACS.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,7 +24,12 @@ jobs:
           REPO: ${{ github.repository }}
         run: |
           # Require an annotated tag — the message becomes the release blurb.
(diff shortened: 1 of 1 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the release-blurb extraction change and have no concerns to add. Interior paragraph breaks are preserved while trailing newlines and whitespace-only messages retain the existing behaviour.

PR #163, opened 2026-06-12
fix(sensor): gate same-day monotonic resets on a plausible post-reset value

dewet22:
Follow-up to #142 — @holdestmade reported the recorder warning recurring on v1.3.0, this time with a *negative* state (−2.2).

**What's actually happening**
The monotonic clamp on `House Consumption Today` treats any same-day drop > 0.5 kWh as a genuine counter reset and accepts it verbatim. That branch exists for an inverter clock lagging HA's midnight by a poll cycle, but it also waves through transient register-skew excursions. I reproduced it on my own install the same day as the report: a one-poll zeroed register read sank the derived consumption from 22.1 to 1.4, the clamp accepted it as a "reset", and the recorder double-counted the recovery as ~21 kWh of new consumption in the long-term statistics. When the skew composition pushes the value below zero — as on holdestmade's install — HA's `total_increasing` guard also fires the warning. The warning is the visible symptom; the silent sum corruption is the real damage.

**The fix**
A same-day drop is now accepted as a reset only when the new value lands in [0, 0.5 kWh] — where a real reset observed within a poll or two always lands (even 20 kW for 30 s accumulates ~0.17 kWh). Anything else holds the previous max until the source recovers, so the recorder never sees a fake reset. Negative values can no longer become the baseline on any path: day flip, the max-clamp branch (which had a falsy-zero bug: a sub-threshold dip off a 0.0 baseline was returned unclamped), or restart restore of a state persisted by pre-fix versions.

The upstream cause — implausible zeroed register reads committing into the plant model — is a givenergy-modbus matter, raised with the library separately. This change makes the one sensor with derived-value exposure robust regardless.

**Verification**
Eight new tests: the #142 reproduction (negative and still-large dips held, recovery not double-counted), negative-baseline flooring on first read/day-flip/restore, the falsy-zero regression, and pins on the behaviours that must not break — midnight reset, clock-lag reset via the gated branch, same-day restore seeding. Full suite green (273), mypy strict clean.

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -1328,6 +1328,8 @@ class GivEnergyInverterSensor(
         self._monotonic_max: float | None = None
         self._monotonic_date: date | None = None
         self._monotonic_last_read: datetime | None = None
+        self._monotonic_reset_pending = False
+        self._monotonic_prior_day_value: float | None = None
         self._source_ir_registers = _source_ir_registers(
             type(coordinator.data.inverter), description.source_field or description.key
         )
@@ -1371,6 +1373,12 @@ class GivEnergyInverterSensor(
             # negative state (#142), which must not re-seed the baseline.
             self._monotonic_max = max(restored, 0.0)
             self._monotonic_date = last_date
+        else:
+            # A prior-day value is no intra-day baseline, but it is the
+            # reference for the first reading of the new day: matching it
+            # means the counter carried over un-reset across a restart
+            # spanning midnight, so the late reset must still be admitted.
+            self._monotonic_prior_day_value = max(restored, 0.0)
 
     @property
     def available(self) -> bool:
@@ -1410,7 +1418,20 @@ class GivEnergyInverterSensor(
                 # negative at any moment, including across the day boundary
                 # (#142). A high carry-over here is normal (inverter clock
                 # lagging HA's midnight); the reset branch below catches the
-                # real reset a poll later.
+                # real reset a poll later. A reset is owed exactly when the
+                # counter did NOT drop at the boundary — judged against the
+                # running max, or after a restart against the restored
+                # prior-day value; only while it is owed may the acceptance
+                # band widen with elapsed time.
+                prior_ref = (
+                    self._monotonic_max
+                    if self._monotonic_max is not None
+                    else self._monotonic_prior_day_value
+                )
+                self._monotonic_reset_pending = (
+                    prior_ref is not None and value >= prior_ref - _MONOTONIC_RESET_THRESHOLD
+                )
+                self._monotonic_prior_day_value = None
                 self._monotonic_max = max(value, 0.0)
                 self._monotonic_date = today
             elif self._monotonic_max is None:
@@ -1424,17 +1445,21 @@ class GivEnergyInverterSensor(
                 # threshold's comment). Anything else is transient register
                 # skew: hold the clamp until the source recovers, so the
                 # recorder never sees a fake reset followed by a
-                # sum-corrupting recovery jump (#142). The ceiling scales
-                # with time since the previous reading so a reset observed
-                # late — long scan interval, or a polling outage spanning
-                # midnight — is not rejected for having accumulated more
-                # than the floor.
+                # sum-corrupting recovery jump (#142). While the midnight
+                # reset is still owed, the ceiling scales with time since
+                # the previous reading so a reset observed late — long scan
+                # interval, or a polling outage after the date flip — is
+                # not rejected for having accumulated more than the floor.
+                # Once it has been observed (or was never owed), the band
+                # stays at the floor: a post-gap daytime sag is skew, not a
+                # reset.
                 ceiling = _MONOTONIC_RESET_THRESHOLD
-                if last_read is not None:
+                if self._monotonic_reset_pending and last_read is not None:
                     elapsed_hours = (now - last_read).total_seconds() / 3600.0
                     ceiling = max(ceiling, _MONOTONIC_RESET_MAX_LOAD_KW * elapsed_hours)
                 if 0.0 <= value <= ceiling:
                     self._monotonic_max = value
+                    self._monotonic_reset_pending = False
             else:
                 self._monotonic_max = max(value, self._monotonic_max)
             return self._monotonic_max
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -1442,17 +1442,29 @@ class GivEnergyInverterSensor(
                 self._monotonic_max = max(value, 0.0)
                 self._monotonic_date = today
             elif self._monotonic_reset_pending is None and value >= 0.0:
-                # First plausible reading of a day whose boundary poll was
-                # skew: make the deferred owed-reset call now, then fall
-                # through to the normal same-day handling on the next poll.
+                # Deferred owed-reset call (the boundary poll was skew).
+                # Three outcomes: the carry-over reappearing means the reset
+                # is still owed; a reading inside the (elapsed-scaled) reset
+                # band means it happened during the skew window; anything in
+                # between is an ambiguous still-large sag — exactly the
+                # shape this clamp rejects — so it neither settles the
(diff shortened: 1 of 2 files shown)


dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -1383,22 +1389,30 @@ def native_value(self) -> Any:
...
+                # below assume a non-None max.
+                self._monotonic_max = max(value, 0.0)
+            elif value < self._monotonic_max - _MONOTONIC_RESET_THRESHOLD:
+                # Large same-day drop: a genuine counter reset, but only when
+                # the new value is a plausible post-reset reading (see the
+                # threshold's comment). Anything else is transient register
+                # skew: hold the clamp until the source recovers, so the
+                # recorder never sees a fake reset followed by a
+                # sum-corrupting recovery jump (#142).
+                if 0.0 <= value <= _MONOTONIC_RESET_THRESHOLD:
```
dewet22-codex: I think the fixed 0.5 kWh post-reset ceiling can reject a genuine late midnight reset. Scan interval is an unrestricted integer in the config flow, and this same monotonic sensor now reads the native three-phase consumption counter too; after a long interval or on a high-load plant, the first post-reset reading can already exceed 0.5 kWh. That leaves yesterday's maximum clamped until today's counter eventually overtakes it, losing most or all of the day's statistics. I think reset plausibility needs to account for elapsed time/cadence (and perhaps proximity to the date flip) rather than requiring every genuine reset to land below a fixed energy value.

dewet22: Addressed in a287aea — the acceptance ceiling now scales with time since the previous reading, max(0.5 kWh, 15 kW × elapsed), so a reset observed late (long scan interval, or an outage after the date flip) passes, while the default 30 s cadence keeps the tight anti-skew band. Covered by three new tests (poll-gap reset, long-interval reset, negative-after-gap still rejected).

custom_components/givenergy_local/sensor.py
```diff
@@ -1383,22 +1389,30 @@ def native_value(self) -> Any:
...
+                    self._monotonic_max = value
             else:
-                self._monotonic_max = max(value, self._monotonic_max or value)
+                self._monotonic_max = max(value, self._monotonic_max)
             return self._monotonic_max
```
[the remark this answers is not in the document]
dewet22: Done in a287aea, though scaling on elapsed time since the previous reading rather than the configured update interval — that also covers a transient poll outage spanning the reset, which a static interval-based threshold would miss. The floor stays at 0.5 kWh and the load bound is 15 kW for three-phase headroom.

dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -1380,25 +1399,44 @@ def available(self) -> bool:
...
+                # recorder never sees a fake reset followed by a
+                # sum-corrupting recovery jump (#142). The ceiling scales
+                # with time since the previous reading so a reset observed
+                # late — long scan interval, or a polling outage spanning
+                # midnight — is not rejected for having accumulated more
+                # than the floor.
+                ceiling = _MONOTONIC_RESET_THRESHOLD
+                if last_read is not None:
+                    elapsed_hours = (now - last_read).total_seconds() / 3600.0
+                    ceiling = max(ceiling, _MONOTONIC_RESET_MAX_LOAD_KW * elapsed_hours)
```
dewet22-codex, line 1490: The elapsed scaling covers late genuine resets, but it currently widens the acceptance band after every polling gap, including ordinary daytime gaps where no reset is possible. For example, after a two-hour outage the ceiling is 30 kWh, so the original still-large skew shape (14.3 -> 12.1) is accepted as a reset and the recovery is double-counted again. I think the elapsed-scaled ceiling needs to apply only while a midnight reset is still pending after a date flip; otherwise the fixed floor should continue rejecting daytime sags. A regression test with a daytime gap followed by a still-large dip would pin that distinction.

dewet22: Addressed in 48db85c — the widened ceiling now applies only while a reset is owed: flagged at the date flip when the counter carries over un-reset (judged against the running max, or the restored prior-day value across a restart spanning midnight) and consumed once the reset is accepted. Ordinary daytime gaps stay on the 0.5 kWh floor, pinned by your daytime-gap-then-sag regression test plus a pending-consumed test.

dewet22-codex:
I’ve started the final review of this PR.

custom_components/givenergy_local/sensor.py
```diff
@@ -1380,25 +1407,61 @@ def available(self) -> bool:
...
+                # counter did NOT drop at the boundary — judged against the
+                # running max, or after a restart against the restored
+                # prior-day value; only while it is owed may the acceptance
+                # band widen with elapsed time.
+                prior_ref = (
+                    self._monotonic_max
+                    if self._monotonic_max is not None
+                    else self._monotonic_prior_day_value
+                )
+                self._monotonic_reset_pending = (
```
dewet22-codex: There is still a boundary-skew sequence that loses the pending reset. If the first new-day poll is a transient negative reading, prior_ref is yesterday's max but this sets pending false and floors the new baseline to zero. When the source recovers to yesterday's still-unreset value on the next poll, the normal max branch raises the baseline again without reconsidering pending; a later genuine reset above the fixed 0.5 kWh floor is then rejected. This is plausible because the surrounding comment explicitly allows negative skew at the day boundary. I think the pending decision needs to survive an implausible boundary reading, or be set when a prior-day carry-over subsequently reappears; a regression sequence such as 14.3 -> date flip/-1.0 -> 14.3 -> gap/1.8 would pin it.

dewet22: Addressed in aa018ef — a negative boundary reading now leaves the owed-reset question undecided (tri-state pending); the first plausible reading of the day makes the call against the retained prior-day reference, so the reappearing carry-over restores the owed state. Your exact sequence (14.3 → flip/−1.0 → 14.3 → gap/1.8) is pinned as a regression test.

dewet22-codex:
I’ve started the final review of this PR.

custom_components/givenergy_local/sensor.py
```diff
@@ -1380,25 +1407,80 @@ def available(self) -> bool:
...
+            elif self._monotonic_reset_pending is None and value >= 0.0:
+                # First plausible reading of a day whose boundary poll was
+                # skew: make the deferred owed-reset call now, then fall
+                # through to the normal same-day handling on the next poll.
+                prior_ref = self._monotonic_prior_day_value
+                self._monotonic_reset_pending = (
+                    prior_ref is not None and value >= prior_ref - _MONOTONIC_RESET_THRESHOLD
+                )
+                self._monotonic_prior_day_value = None
+                self._monotonic_max = max(
```
dewet22-codex: The negative-boundary sequence is covered now, but the first non-negative reading can still be an ambiguous still-large skew sag rather than recovered carry-over or a plausible reset. For example: 14.3 -> date flip/-1.0 -> 12.1 -> 14.3 -> gap/1.8. This branch decides pending=false on 12.1 and exposes it as the new max; carry-over recovery then raises the max to 14.3, and the later genuine 1.8 reset is rejected by the fixed floor. I think the deferred decision needs to remain undecided for a reading that is neither within the reset band nor close to the prior-day reference, so another plausible poll can settle it. A regression using that sequence would cover the same still-large skew shape the PR is intended to reject.

dewet22: Addressed in d381e8a — the deferred call now has three outcomes: carry-over reappearing keeps the reset owed; a reading inside the elapsed-scaled reset band counts it as seen; the ambiguous middle holds the current max, stays undecided, and is not exposed. Your sequence (14.3 → flip/−1.0 → 12.1 → 14.3 → gap/1.8) is pinned, plus the plausible-reset arm.

dewet22-codex:
I’ve started the final review of this PR.

dewet22-codex (approved):
I’ve completed the final review and have no concerns to add. The deferred reset decision now holds ambiguous boundary-skew readings, admits late genuine resets, and keeps ordinary daytime gaps on the tight floor.

Issue #162, opened 2026-06-12
migrate_from_givtcp: re-base sum statistics at the join, optionally rebuild implausible source sums


dewet22:
## Background

Running `scripts/migrate_from_givtcp.py` on my install backported GivTCP's long-term statistics onto this integration's sensors (7 June 2026, ~15:00 BST). The mean/min/max columns came across cleanly — History views and anything mean-derived look right — but the `sum` column ended up wrong in two distinct ways, which took some untangling on a year-scale chart before the causes were clear.

## Observed problems

**1. The migration itself creates a seam at the join.** The backported rows carry GivTCP's cumulative sum (~2364 kWh at the join), while the integration's own rows continue from their original near-zero baseline. The hour the two meet shows a single `change` of **−2362.3 kWh**, which wrecks any chart computing sums/changes across it (a year of weekly PV bars rendered as one giant negative spike).

**2. Broken source sums are copied verbatim.** GivTCP's own `sum` column was structurally broken on this install until ~Dec 2025: summer-2025 months show net changes of ±15 kWh where ~600 kWh of generation actually occurred, including outright negative months (Jul −14.6, Sep −8.8) and a −165.4 kWh step on 19 Nov 2025 (with no rows at all for 1–18 Nov). I verified the source sensor (`sensor.givtcp_*_pv_energy_today_kwh`) shows the identical pattern, so this is inherited rather than introduced — but the migration currently propagates it silently. A PV generation counter should never produce a negative change.

## Intended fixes

- **Re-base sums at the join** so the spliced range is continuous with the target's existing rows: apply a constant offset to the backported rows (or to the subsequent native rows) such that no artificial step appears at either end of the spliced range.
- **Optional `--rebuild-sums`**: where the source `sum` column is implausible (negative changes on a monotonic quantity, long flat stretches contradicted by the mean column), reconstruct the sum from the `state`/mean columns instead of copying it.
- **Post-migration validation pass**: scan the result for negative hourly/daily changes on `total_increasing`-style statistics and report them, so a seam can't sit undiscovered until someone plots a year.

## Workaround used meanwhile

The two large seams are fixable by hand in Developer Tools → Statistics → *adjust sum* (set the bad hour's change to 0): 7 Jun 2026 15:00 BST (−2362.3) and 19 Nov 2025 16:00 GMT (−165.4). The inherited 2025 undercount remains until a re-migration with rebuilt sums.

[...]

PR #168, opened 2026-06-15
feat(migrate): rebuild GivTCP sums, validate, and back-port mean series (#162)

dewet22:
## Summary

Reworks `scripts/migrate_from_givtcp.py` to tackle the sum-statistics corruption described in #162.

- **Rebuild sums by default.** Rather than copying GivTCP's `sum` column (structurally broken on my install — negative months, midnight-negative backfill), the script now concatenates the `state` timeline across the cut-over and walks it once into a single continuous sum, eliminating the join seam by construction. The walk is plausibility-guarded and reset-aware — a decrease only counts as a genuine reset at a counter's natural boundary (daily counters at local midnight, DST-aware; lifetime counters never); off-boundary drops and over-ceiling spikes hold the last-good **state and sum**. `--trust-source-sums` restores the old copy-and-rebase path.
- **Fail-closed plausibility ceiling.** The adaptive per-entity ceiling returns no bound when there's too little clean data to estimate one; combined with an optional `--max-kw` hard cap (your inverter's peak output). When neither yields a bound, the entity is refused (`insufficient_data`, no clear+import) rather than imported unguarded.
- **Post-migration validation report** — read-only, runs in both dry-run and apply; flags residual implausible hours, fake-reset shapes, duplicate series, and gaps (gaps informational). `--repair-residue` opts into targeted re-import for sum entities that still look wrong.

## Notes / caveats

- The **mean/power/SOC/temperature back-port was dropped** from this PR: checking its mappings against a live registry (during Codex review) found several wrong/unsafe (e.g. GivTCP `grid_power` ≠ GE signed-net `grid_power`), and the clear-before-import would overwrite real history. It's split to #169 to be redone with verified mappings + tests.
- The validation/repair WebSocket paths have unit coverage on their pure pieces but haven't been exercised end-to-end against a live HA instance.
- Includes the design spec and implementation plan under `docs/superpowers/`.

## Test plan

- [x] `uv run pytest` — 320 passing
- [x] ruff check + format clean
- [x] Codex + CodeRabbit + Gemini review threads addressed
- [ ] Dry-run against a live HA recorder; eyeball the validation report

## Summary by CodeRabbit

* **New Features**
  * Energy sensor migrations now rebuild continuous sums instead of rebasing, eliminating dashboard discontinuities at the cut-over point.
  * Added post-migration validation flagging implausible values, duplicates, and data gaps.
  * Optional residue repair mode re-imports entities with persistent data quality issues.
  * New CLI flags: `--trust-source-sums` (legacy behaviour), `--repair-residue` (targeted fixes), and `--max-kw` (plausibility bounds).

* **Documentation**
  * Updated migration design notes and implementation specifications covering rebuild strategy, validation, and repair workflows.

* **Tests**
  * Comprehensive test suite validating rebuild logic, validation checks, and edge case handling.


dewet22-codex:
Starting a migration-safety review round now.

scripts/migrate_from_givtcp.py
```diff
@@ -154,6 +185,25 @@
 # is omitted rather than mis-migrated. (Was added in #126, reverted here.)
 BATTERY_PAIRS: list[tuple[str, str, str, str]] = []
 
+# (givtcp_suffix, ge_suffix, description) — mean-type series (power/SOC/temp).
+# Straight mean/min/max copy; no sum, no rebase, no plausibility.
+# NOTE: these suffixes are provisional and must be verified against a live
+# GivTCP + givenergy_local registry before relying on them in anger.
+MEAN_PAIRS: list[tuple[str, str, str]] = [
+    ("pv_power", "pv_power", "PV power"),
+    ("grid_power", "grid_power", "Grid power (signed)"),
```
dewet22-codex: [P1] Do not enable these provisional mean mappings by default. This specific pair is already contradicted by this repository's own migration catalogue: GivTCP `grid_power` maps to givenergy_local `grid_power_phase_1`, while givenergy_local `grid_power` is the signed net-meter signal corresponding to GivTCP `export_power`. Because mean migration is default-on and clears the recognised target before import, `--apply` will replace the signed-net `grid_power` history with a different phase-power series. The same table also names unresolved/provisional targets such as `house_consumption` and per-battery `temperature`. Please verify and correct every pair against live registries/semantics, with tests pinning the mappings, or make this destructive back-port explicitly opt-in until that verification is complete.

scripts/migrate_from_givtcp.py
```diff
@@ -247,6 +297,144 @@ def build_entity_id_resolver(
...
+
+    Uses median + K * 1.4826 * MAD over the positive, finite deltas. Both the
+    median and the MAD are resistant to a handful of giant outliers, so the
+    bound reflects genuine hourly behaviour even on a heavily corrupted series.
+    Returns +inf when there is nothing positive to anchor on (caller then can't
+    guard, and the walk accepts all non-negative deltas).
+    """
+    pos = sorted(d for d in deltas if d is not None and d > 0)
+    if not pos:
+        return float("inf")
```
dewet22-codex: [P1] Fail closed when there is insufficient clean data to estimate a ceiling. Returning infinity disables the only spike guard for constant/empty-positive-delta series, and a sparse series is also unsafe: for positive deltas `[1, 9900]`, the median/MAD calculation produces a ceiling above 60,000, so the 9,900 spike is accepted. Post-migration validation derives the same inflated ceiling and therefore blesses the corruption. Since rebuild is the destructive default, require a defensible minimum sample/distribution, apply a domain-safe upper bound, or skip/refuse that entity rather than importing unguarded sums.

scripts/migrate_from_givtcp.py
```diff
@@ -247,6 +297,144 @@ def build_entity_id_resolver(
...
+            out.append(r)
+            continue
+        delta = state - prev_state
+        if 0 <= delta <= ceiling:
+            running += delta
+            prev_state = state
+        elif delta < 0 and _is_reset_boundary(r["start"], reset_class, tz, midnight_tol_hours):
+            running += state
+            prev_state = state
+        # else: corruption (off-boundary drop) or spike (> ceiling) -> hold last-good.
```
dewet22-codex: [P1] Actually hold the last-good state, not only the sum. `r` remains a copy of the corrupt row here, so an off-boundary zero or giant spike is imported unchanged in `state` even though its `sum` is held. That leaves the migrated state timeline corrupted, makes the post-migration fake-reset check continue to fail, and `--repair-residue` cannot heal it because repair is queued only for implausible sum changes and then repeats this same walk. Set the rejected row's state to the trusted state (or omit it with a deliberate policy) and add an end-to-end assertion that full validation is clean after rebuilding the documented corruption.

scripts/migrate_from_givtcp.py
```diff
@@ -636,20 +1165,56 @@ def _build_plan(
...
+    plan: list[tuple[str, str, str, str, bool]] = []
+    for sn in inv_serials:
+        for gt_sfx, ge_sfx, desc in MEAN_PAIRS:
+            givtcp_id = f"sensor.givtcp_{sn}_{gt_sfx}"
+            canonical = f"sensor.givenergy_inverter_{sn}_{ge_sfx}"
+            resolved = canonical in canonical_to_actual
+            ge_id = canonical_to_actual.get(canonical, canonical)
+            unit = meta_by_id.get(ge_id, {}).get("unit_of_measurement") or "W"
+            plan.append((givtcp_id, ge_id, desc, unit, resolved))
+    for sn in batt_serials:
```
dewet22-codex: [P2] The per-battery mean plan is unreachable with the current discovery path. `batt_serials` is detected from `_BATT_DETECT` matching `battery_cycles`, but `run()` builds `all_ids` from `list_statistic_ids()` with its default `statistic_type="sum"`; `battery_cycles` is explicitly documented here as a mean statistic, so it will not be in that list. Consequently SOC/temperature back-port silently skips every battery. Discover battery serials from the mean statistic IDs or the registry, and cover the full run-plan construction in a regression test.

dewet22-codex (changes requested):

scripts/migrate_from_givtcp.py
```diff
@@ -579,6 +952,160 @@ async def migrate_entity(
...
+    series read back are the *current* (unmigrated) HA series, not a
+    post-migration result.  When ``repair`` is set (only when --apply and
+    --repair-residue are both given), entities with residual implausible hours
+    have their rebuilt series cleared and re-imported; ``units_by_id`` supplies
+    the unit of measurement for the re-import metadata.  Only sum entities
+    (those present in ``units_by_id``) are eligible for repair — mean entities
+    are never touched.
+    Returns the validation exit code from ``format_validation_report``.
+    """
+    units_by_id = units_by_id or {}
```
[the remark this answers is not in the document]
dewet22: Done in 2621168 — `run_validation` now takes a `reset_classes` mapping built from the plan, so repair uses the plan's authoritative cadence rather than re-inferring from the (possibly renamed) entity id.

scripts/migrate_from_givtcp.py
```diff
@@ -579,6 +952,160 @@ async def migrate_entity(
...
+        print(
+            f"  Repairing {len(to_repair)} entit{'y' if len(to_repair) == 1 else 'ies'} with"
+            " residual implausible hours …"
+        )
+        for ge_id in to_repair:
+            rows = series_by_id[ge_id]
+            ceiling = adaptive_ceiling(_state_deltas(rows))
+            # classify_entity keys off the suffix via endswith, so the full
+            # resolved ge_id (…_today / …_total / …_this_year) classifies correctly.
+            rebuilt = rebuild_sum_walk(rows, classify_entity(ge_id), ceiling, tz)
```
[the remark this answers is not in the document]
dewet22: Done — repair now resolves the class via `_repair_reset_class(ge_id, reset_classes)` (extracted in b86baef), falling back to `classify_entity` only when the entity isn't in the plan. A unit test pins the renamed-entity case (plan says DAILY where the suffix would infer LIFETIME).

scripts/migrate_from_givtcp.py
```diff
@@ -830,23 +1398,65 @@ def resolve(eid: str) -> str:
...
+    # Read-only validation of the resulting sum series, in both dry-run and apply.
+    # Residue repair only fires when the user opted in with --apply --repair-residue.
+    validation_exit = await run_validation(
+        ws,
+        results,
+        tz,
+        units_by_id=units_by_id,
+        repair=applying and args.repair_residue,
+        applied=applying,
+    )
```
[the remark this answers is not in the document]
dewet22: Done in 2621168 — the mapping is built as `{ge_id: reset_class for (_, ge_id, _, _, _, reset_class, _) in plan}` and passed into `run_validation`.

scripts/migrate_from_givtcp.py
dewet22: Verified against the live registry — you're right, and on more pairs than `grid_power` (`load_power`→`house_consumption` and the per-battery `temperature` target are also wrong). Rather than ship half-verified destructive mappings, I've dropped the mean back-port from this PR in 98747b1 and will redo it properly in #169 with per-pair live-registry + sign-convention verification and tests.

scripts/migrate_from_givtcp.py
dewet22: Fixed in 417b733. `adaptive_ceiling` now returns None below a minimum sample count (and on empty — no more inf); a new `effective_ceiling` combines it with an optional `--max-kw` hard cap (min of the two). When neither yields a bound, `migrate_entity` refuses the entity (`insufficient_data`, no clear+import) and `run_validation` skips the plausibility checks rather than re-deriving and re-blessing. The [1, 9900] sparse case is now refused, not imported.

scripts/migrate_from_givtcp.py
dewet22: Fixed in 417b733 — the hold branch now sets the rejected row's `state` to last-good as well as holding the sum, so the imported state timeline is clean (prev_state isn't advanced, so the recovery is measured against the last trusted reading). The acceptance test now asserts `find_fake_reset_shapes(out, ceiling) == []` and that no corrupt state survives after rebuilding the documented corruption.

scripts/migrate_from_givtcp.py
dewet22: Resolved by removing the mean back-port from this PR (98747b1) — there's no battery-mean discovery path anymore. Correct serial discovery (not from the sum-only statistic list) plus a run-plan regression test are in the #169 acceptance criteria.

dewet22-codex:
Starting a migration-safety re-review round now.

scripts/migrate_from_givtcp.py
```diff
@@ -896,6 +1444,33 @@ def main() -> None:
...
+            "rebuilding sums from state. Use only if your GivTCP sums are known-good."
+        ),
+    )
+    p.add_argument(
+        "--max-kw",
+        type=float,
+        default=None,
+        metavar="KW",
+        help=(
+            "Your inverter's peak output in kW. Hourly energy deltas above this "
```
dewet22-codex: [P1] Do not use inverter peak output as the global ceiling. `args.max_kw` is passed to every entity, including grid import, house consumption, charge-from-grid, and battery throughput, whose legitimate hourly delta can exceed inverter output. In the sparse-data case this value becomes the only ceiling, so such a legitimate delta is held/flattened during import; validation then reuses the same ceiling and can report the damaged series as clean. Please define this as a positive maximum legitimate hourly delta covering every migrated counter (and reject non-positive values), or support an appropriate per-entity bound instead.

dewet22-codex (changes requested):

scripts/migrate_from_givtcp.py
dewet22: Fixed in 7bacb41. `--max-kw` is no longer framed as inverter output — it's now defined as an upper bound on a plausible hourly energy change (kW = kWh/h) applied to every counter, with the help text noting grid import and battery charging can exceed PV output, so it's set above the largest legitimate hourly delta any counter sees. Non-positive values are rejected at parse time (`_positive_float` → ArgumentTypeError), with unit tests. Kept it a single global bound rather than per-entity, per your first option.

dewet22-codex:
Starting a migration-safety re-review round now.

scripts/migrate_from_givtcp.py
```diff
@@ -830,23 +1346,63 @@ def resolve(eid: str) -> str:
...
     await ws.close()
 
-    return _print_summary(results, applying)
+    return max(summary_code, validation_exit)
+
+
+def _positive_float(value: str) -> float:
+    """argparse type: a strictly positive float (for --max-kw)."""
+    f = float(value)
+    if f <= 0:
```
dewet22-codex: [P1] Reject non-finite caps as well. Python accepts `--max-kw nan` and `--max-kw inf`, and neither satisfies `f <= 0`. With sparse data, `nan` becomes the ceiling and causes every nonnegative delta comparison to fail, silently flattening the imported series; `inf` restores the unguarded import this fail-closed path was added to prevent. Please require `math.isfinite(f) and f > 0`, with coverage for both values.

dewet22-codex (changes requested):

scripts/migrate_from_givtcp.py
dewet22: Fixed in 3398e7c — `_positive_float` now requires `math.isfinite(f) and f > 0`, so `nan`, `inf`, and `-inf` are all rejected at parse time (with test coverage for each). Good catch on nan slipping the `<= 0` check and silently flattening, and inf restoring the unguarded import.

dewet22-codex:
Starting a migration-safety re-review round now.

scripts/migrate_from_givtcp.py
```diff
@@ -713,8 +1219,16 @@ def _print_summary(results: list[MigrationResult], applying: bool) -> int:
                 "Skipped.",
                 file=sys.stderr,
             )
+        elif r.status == "insufficient_data":
+            print(
+                f"\n  ⚠️  {r.description}: too little clean history to estimate a "
+                "plausibility ceiling, and no --max-kw cap was given — skipped rather "
+                "than import an unguarded sum. Re-run with --max-kw <your inverter's "
```
dewet22-codex: [P1] Keep this recovery guidance consistent with the corrected cap semantics. This is the message sparse-data users will actually see, and it still instructs them to use inverter peak output; for grid import, house consumption, or battery charging that can be below legitimate hourly movement and silently flatten the imported series. Please tell them to choose a value above the largest legitimate hourly delta across every migrated counter, as the CLI help now does.

dewet22-codex (changes requested):

scripts/migrate_from_givtcp.py
dewet22: Fixed in 0646024 — the insufficient_data recovery hint now tells users to pick a value above the largest legitimate hourly delta across all counters (noting grid import and battery charging can exceed PV output), matching the corrected --max-kw help. No more 'inverter peak output' wording.

dewet22-codex:
Starting a migration-safety re-review round now.

dewet22-codex (approved):
Migration-safety re-review complete; no remaining concerns.

dewet22-codex:
Final migration-safety re-review complete. All prior concerns are addressed; approved.

PR #170, opened 2026-06-15
fix(migrate): p99-based plausibility ceiling (median+MAD over-flattened genuine peaks) (#162)

dewet22:
## Summary

Follow-up to #168. A live dry-run showed `adaptive_ceiling` (median + 8·MAD) collapses to ~3 kWh/h on high-variance counters (solar generation, house consumption) — the median sits at the near-zero overnight floor, so genuine daytime/evening peaks look like outliers. That over-flagged hundreds of genuine hours, and since the rebuild walk shares the ceiling, `--apply` would have **flattened genuine generation/consumption peaks**.

Replaces it with a **99th-percentile × 1.5** ceiling. Genuine hourly deltas are bounded by hardware (PV inverter clip ~6 kWh/h; battery subsystem ~2.5 kWh/h) and corruption is a <1% tail (27,000+ fake-reset spikes), so p99 tracks each counter's real physical limit without being pulled up by the rare fakes; ×1.5 adds headroom. Fail-closed below 24 samples is unchanged.

## Evidence (live dry-run, before → after)

- house_consumption_today: 619 → 69 flags (smallest +3.1 → +12.3 kWh/h)
- pv_generation_today: 284 → 42 (now only ≥+7.5, above the ~6 kWh/h PV clip)
- grid_export_today: 51 → 4
- Battery counters correctly *tightened* — `battery_charge` now catches values above the ~2.5 kWh/h subsystem limit that median+MAD missed.
- Genuine fake-reset spikes (+27k / +173 / +84) still flag.

## Test plan

- [x] `uv run pytest` — 325 passing; new regression test pins the diurnal failure mode
- [x] Re-ran the live dry-run to confirm the collapse above

## Release Notes

* **Bug Fixes**
  * Improved per-hour sum plausibility ceiling calculation using a more accurate high-percentile approach.
  * Added mandatory safety requirement: the migration script now enforces a required flag when performing destructive operations.
  * Updated command-line help text to clarify the mandatory flag requirement for safety-critical operations.

* **Tests**
  * Expanded test coverage for plausibility ceiling validation and safety enforcement checks.

### DIFF
--- a/scripts/migrate_from_givtcp.py
+++ b/scripts/migrate_from_givtcp.py
@@ -32,7 +32,8 @@ Typical workflow:
         --ha-url http://homeassistant.local:8123 \\
         --token eyJhbGc... \\
         --cutover 2024-10-31 \\
-        --apply
+        --apply \\
+        --max-kw 10
 
 The cut-over date is the boundary between GivTCP and givenergy_local history.
 GivTCP data strictly before midnight on that date is migrated; givenergy_local
@@ -302,11 +303,12 @@ _CEILING_FACTOR = 1.5
 
 
 def _percentile(sorted_vals: list[float], p: float) -> float:
-    """Linear-interpolated percentile of an ascending-sorted, non-empty list."""
-    k = (len(sorted_vals) - 1) * p / 100.0
-    f = int(k)
-    c = min(f + 1, len(sorted_vals) - 1)
-    return sorted_vals[f] + (sorted_vals[c] - sorted_vals[f]) * (k - f)
+    """Nearest-rank (floor-index) percentile of an ascending-sorted, non-empty
+    list. Uses the floor index rather than linear interpolation so a single
+    order-of-magnitude outlier near the top cannot drag the percentile up toward
+    itself on small samples (which would let the spike bless itself through the
+    ceiling)."""
+    return sorted_vals[int((len(sorted_vals) - 1) * p / 100.0)]
 
 
 def adaptive_ceiling(deltas: list[float | None]) -> float | None:
@@ -324,6 +326,11 @@ def adaptive_ceiling(deltas: list[float | None]) -> float | None:
     if len(pos) < _CEILING_MIN_SAMPLES:
         return None
     return _percentile(pos, _CEILING_PERCENTILE) * _CEILING_FACTOR
+
+
+def _apply_requires_cap(apply: bool, max_kw: float | None) -> bool:
+    """True when an --apply run is missing the required --max-kw bound."""
+    return apply and max_kw is None
 
 
 def effective_ceiling(adaptive: float | None, cap: float | None) -> float | None:
@@ -1239,6 +1246,17 @@ async def run(args: argparse.Namespace) -> int:
     print(f"Mode           : {mode}")
     print()
 
+    if _apply_requires_cap(applying, args.max_kw):
+        print(
+            "\n  ✋ --apply requires --max-kw. The destructive rebuild needs a "
+            "trusted upper bound on a plausible hourly delta (kW) — the adaptive "
+            "ceiling alone can be skewed by heavy corruption. Run a dry-run first "
+            "to see the plan, then re-run with --apply --max-kw <kW above your "
+            "largest legitimate hourly delta across all counters>.",
+            file=sys.stderr,
+        )
+        return 2
+
     print("Connecting …", end=" ", flush=True)
     ws = HAWebSocket(args.ha_url, args.token)
     try:
@@ -1441,7 +1459,8 @@ def main() -> None:
         action="store_true",
         help=(
             "Write changes to Home Assistant. Without this flag the script runs in "
-            "dry-run mode and prints a preview without modifying anything."
+            "dry-run mode and prints a preview without modifying anything. "
+            "Requires --max-kw."
         ),
     )
     p.add_argument(
@@ -1472,7 +1491,7 @@ def main() -> None:
             "legitimate hourly delta any of your counters can see — grid import "
             "and battery charging can exceed your inverter's PV output. Must be "
             "positive. Caps the adaptive ceiling and lets entities with too little "
-            "history to self-estimate still rebuild safely."
+            "history to self-estimate still rebuild safely. Required with --apply."
         ),
     )
     p.add_argument(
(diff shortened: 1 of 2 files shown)


dewet22-codex:
Starting a code review round now.

scripts/migrate_from_givtcp.py
```diff
@@ -297,35 +296,34 @@ def classify_entity(ge_suffix: str) -> ResetClass:
...
+    guessing a bound from too little data.
     """
     pos = sorted(d for d in deltas if d is not None and d > 0)
     if len(pos) < _CEILING_MIN_SAMPLES:
         return None
-    median = statistics.median(pos)
-    mad = statistics.median([abs(d - median) for d in pos])
-    spread = mad if mad > 0 else median
-    return median + _CEILING_MAD_K * 1.4826 * spread
+    return _percentile(pos, _CEILING_PERCENTILE) * _CEILING_FACTOR
```
dewet22-codex, line 330: [P1] Do not make destructive rebuild safety depend on corruption staying below 1%. Even after replacing interpolation with a floor-index percentile, p99 becomes a corrupt value as soon as roughly 1% of an entity's positive deltas are fake (for example, 2 recurring fake-reset spikes among 100 deltas). Multiplying that value by 1.5 then blesses those spikes, and the rebuild imports them because it shares this ceiling. The new regression only covers 2 fakes among 282 values, which bakes in the unverified <1% assumption. Please use an estimator/guard that remains bounded when the corrupt tail exceeds this threshold, or require a trusted hard cap before applying when the data cannot establish one safely.

dewet22-codex (changes requested):

scripts/migrate_from_givtcp.py
```diff
@@ -297,35 +296,34 @@ def classify_entity(ge_suffix: str) -> ResetClass:
...
+_CEILING_PERCENTILE = 99.0
+_CEILING_FACTOR = 1.5
+
+
+def _percentile(sorted_vals: list[float], p: float) -> float:
+    """Linear-interpolated percentile of an ascending-sorted, non-empty list."""
+    k = (len(sorted_vals) - 1) * p / 100.0
+    f = int(k)
+    c = min(f + 1, len(sorted_vals) - 1)
+    return sorted_vals[f] + (sorted_vals[c] - sorted_vals[f]) * (k - f)
```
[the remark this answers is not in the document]
dewet22: Fixed in df502a9 — `_percentile` now uses the floor index (nearest-rank) rather than linear interpolation, so a single order-of-magnitude outlier at the top can't drag p99 toward itself on small samples. Added a regression pinning your exact N=24 single-outlier case (ceiling lands ~9, the 27000 is rejected).

tests/test_migrate_stats_repair.py
```diff
@@ -47,11 +47,30 @@ def test_adaptive_ceiling_no_positive_deltas_returns_none():
...
+    # Sparse data: too few positive deltas. Returning a value here would let the
+    # p99 estimate bless an order-of-magnitude spike.
     assert _MOD.adaptive_ceiling([1.0, 9900.0]) is None
 
 
+def test_percentile_interpolates():
+    vals = [1.0, 2.0, 3.0, 4.0]
+    assert _MOD._percentile(vals, 50) == 2.5
+    assert _MOD._percentile(vals, 0) == 1.0
+    assert _MOD._percentile(vals, 100) == 4.0
```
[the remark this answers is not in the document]
dewet22: Done — updated to `test_percentile_floor` asserting the floor-index result (p50 of [1,2,3,4] = 2.0).

scripts/migrate_from_givtcp.py
dewet22: Agreed — the destructive path no longer relies on the corruption-fraction assumption. As of df502a9, `--apply` now requires `--max-kw`: with a user-supplied cap, `min(adaptive, cap) <= cap` rejects spikes above the stated max regardless of how corrupt the data is. `run()` fails fast (exit 2) if `--apply` is given without `--max-kw`; dry-run stays adaptive-only. The p99 estimator remains for the dry-run report only.

scripts/migrate_from_givtcp.py
```diff
@@ -297,35 +297,40 @@ def classify_entity(ge_suffix: str) -> ResetClass:
...
+    guessing a bound from too little data.
     """
     pos = sorted(d for d in deltas if d is not None and d > 0)
     if len(pos) < _CEILING_MIN_SAMPLES:
         return None
-    median = statistics.median(pos)
-    mad = statistics.median([abs(d - median) for d in pos])
-    spread = mad if mad > 0 else median
-    return median + _CEILING_MAD_K * 1.4826 * spread
+    return _percentile(pos, _CEILING_PERCENTILE) * _CEILING_FACTOR
```
dewet22-codex, line 330: [P1] Use the trusted cap as the destructive rebuild ceiling, rather than still taking `min(adaptive, cap)`. Requiring `--max-kw` only bounds an adaptive estimate that is too high; it does not prevent p99 from being too low. A legitimate high-load/grid-import hour occurring in less than roughly 1% of positive deltas can sit above `p99 * 1.5`, so `--apply --max-kw <correct higher bound>` still chooses the lower adaptive value and silently flattens that genuine peak. This is the same class of destructive underestimation that prompted this PR. Keep p99 for dry-run/reporting if useful, but once apply requires a user-declared maximum legitimate delta, the rebuild should use that trusted bound (or otherwise prove the adaptive lower bound cannot reject legitimate peaks).

dewet22-codex (changes requested):

scripts/migrate_from_givtcp.py
dewet22: Agreed, and verified against the live data: house_consumption p99x1.5≈12 but the genuine max is 19.8 kWh/h (EV), so min(adaptive,cap) would flatten that even with --max-kw 20 set. Fixed in 76dc27c — effective_ceiling now takes the cap with PRECEDENCE (not min): under --apply the user-declared --max-kw is used directly as the rebuild bound, so deltas up to it are preserved and only those above it are rejected. The adaptive p99 remains the dry-run/report fallback when no cap is given.

dewet22-codex (approved):
Re-review complete. The trusted --max-kw cap now governs the destructive path end to end, and all prior concerns are addressed. Approved.

Issue #179, opened 2026-06-17
Support HV battery stacks (plant.hv_stacks): per-stack SOC/SOH/voltage/temperature/module count


dewet22:
The integration currently surfaces battery data only for low-voltage batteries (`coordinator.data.batteries`) and AIO modules. **HV battery stacks (`plant.hv_stacks` — `Bcu`/`Bmu`) are not exposed at all** — they exist in the integration only as `bcu_stacks` topology metadata for device-loss detection (coordinator.py), with no entities created.

Surfaced by #174 — a GIV-3HY-11, the first three-phase + HV-battery system reported. On that system `coordinator.data.batteries` is empty, so **no per-pack battery entities are created**, and the cluster data the library *does* decode goes unsurfaced entirely.

The library's HV stack model decodes cluster-level fields cleanly (validated against #174's capture): `number_of_modules`, `battery_soc_max` / `battery_soc_min`, `battery_soh`, `battery_voltage`, `cluster_cell_temperature`. Proposed: add an HV-stack platform path (parallel to the LV battery and AIO loops) that creates per-stack entities from `plant.hv_stacks`.

**Dependencies / caveats:**
- **Cluster temperature scale** (`cluster_cell_temperature`, IR68) is unconfirmed — likely deci (→ ~7 °C); tracked upstream as givenergy-modbus#265. Wire it expecting the scale to settle there.
- **Per-cell BMU temps** are not yet trustworthy (module-0 stride overlaps the BCU block) — pending an HV per-module layout map from a targeted probe (givenergy-modbus#265). Don't surface per-cell HV temps until that lands.
- The inverter-level Battery Temperature on an HV unit could be sourced from `Bcu.cluster_cell_temperature` once this path exists (relevant to #174).

Context: #174, givenergy-modbus#265.

dewet22:
Worth extending this slightly: as of **givenergy-modbus 2.7.0**, the HV decode now reaches per-**module** granularity, not just per-stack — and it's wire-confirmed against real hardware (the 6-module GIV-3HY-11 stack from #174).

The library exposes it through the same `plant.hv_stacks` entry point this issue already tracks:

- `plant.hv_stacks[]` → `HvStack`, each with `.bcu` (the per-stack cluster aggregate — SOC/SOH/voltage/temperature/module count, the original scope here) **and** `.bmus[]`
[...]

Issue #174, opened 2026-06-16
3 phase hybrid installation issues


lamztib:
I just enabled this integration to my system:
Hybrid (3-phase)
Inverter Model = GIV-3HY-11
Software Version = DA0.011-DD0.011-A0.011
Battery Capacity = 52 Ah | 20.37 kWh | (6 Packs)

I see some entities, but they seem not to refresh all. Some of them have data, like SOC, but some of them do not update. These are some ones i.e. "Battery Power", "Temperature", Throughput.

Would like to make a capture, but i dont find this "Services" part under the Developer tools.

Also with the dashboard instructions, i get the error/message:
Error loading the dashboard strategy:
Error: Timeout waiting for strategy element ll-strategy-dashboard-givenergy to be registered

dewet22:
Thanks for the detailed report — model, firmware and pack count were exactly what I needed.

**Finding the capture:** recent Home Assistant renamed *Developer Tools → Services* to *Developer Tools → Actions*. It's there under that name — search for `givenergy_local.capture_frames` — and there's also a one-tap "Capture Debug Frames" button on the integration's bundled Diagnostics dashboard view. (The README still said "Services"; I've fixed that.)

If you'd rather not use the Home Assistant UI, the companion CLI can do it without a full install — you just need [uv](https://docs.astral.sh/uv/) and your inverter's IP address:

```
uvx givenergy-cli --host 192.168.x.x capture --output frames.log --duration 60
```

For this particular question an `export` is even more helpful, since it dumps every register from every device in one shot:

```
uvx givenergy-cli --host 192.168.x.x export -o plant.json
```

(Heads-up: unlike the in-app capture, these CLI files aren't auto-redacted — they'll contain your device serial numbers, so you may want to trim those before attaching anything publicly.)

**Battery sensors not refreshing:** I've traced this. On three-phase inverters, Battery Power, Battery Temperature and Battery Throughput are inherited from the single-phase register layout, and your firmware doesn't populate those addresses — so they decode once and then sit still, while SOC (which has its own three-phase register) keeps updating. The correct three-phase readings live at different registers, and I'm preparing a fix to source them from there. Yours is the first three-phase system I've had data from, so a capture or export would genuinely help me confirm the exact registers before I rely on them.

[...]

PR #180, opened 2026-06-17
fix: adopt givenergy-modbus 2.3.3 + gate Battery Temperature on three-phase (#174)

dewet22:
Adopts givenergy-modbus 2.3.3 and completes the three-phase inverter-level battery sensors for #174.

**The dependency bump (auto-generated)** corrects three of the four 3ph battery sensors at the library level, so they need no integration change:
- **Battery Power** and **Battery Throughput** now read the native three-phase registers (the #262 facade) instead of the frozen inherited single-phase ones.
- **Battery Current**'s 10× scale is fixed (#264).

**The added commit** handles the fourth: **Battery Temperature** (`t_battery`) has no inverter-level three-phase source — on 3ph it inherits a single-phase register the firmware never populates, so it reads frozen. It's now `single_phase_only`, dropped on three-phase rather than surfacing a stale value. The real 3ph battery temperature will come from the HV cluster (`Bcu.cluster_cell_temperature`) once HV-stack support lands (#179). Single-phase users are unaffected (the gate fires only on three-phase).

Addresses the battery-sensor part of #174; the dashboard and per-pack HV work are tracked separately (#179).

### DIFF
--- a/custom_components/givenergy_local/manifest.json
+++ b/custom_components/givenergy_local/manifest.json
@@ -13,7 +13,7 @@
   "iot_class": "local_polling",
   "issue_tracker": "https://github.com/dewet22/givenergy-hass/issues",
   "requirements": [
-    "givenergy-modbus>=2.3.2,<3.0.0"
+    "givenergy-modbus>=2.3.3,<3.0.0"
   ],
   "version": "1.1.0"
 }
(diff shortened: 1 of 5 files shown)


dewet22-codex:
Starting review.

dewet22-codex (approved):
Looks good. The `givenergy-modbus` 2.3.3 requirement is updated consistently in the manifest, project metadata, and lockfile. The integration still uses the deprecated `set_charge_target` alias, but 2.3.3 keeps it backwards-compatible, so I do not see a blocker for this dependency bump.

Issue #176, opened 2026-06-16
Coordinator serves a device's frozen data indefinitely when its reads keep partially failing (no auto-recovery, no staleness flag)


dewet22:
**Summary**

When one device's register reads fail persistently — a `RefreshPartiallySucceeded` on every tick — but the device is still *present* in the topology, the coordinator serves that device's frozen last-known values indefinitely, with no automatic recovery and, for battery sensors, no staleness indication. The result is confidently-wrong data until the user manually reloads the config entry.

**Reported behaviour**

A user's primary battery SOC read 19% for 10+ hours while the real value was ~89%; reloading the entry restored correct reporting. They report it recurring ("the integration seems to go to sleep") — spotted when a value looks off and the entity shows it hasn't updated in a long time, always cleared by a reload.

**Mechanism (hypothesis, pending confirmation)** — three compounding layers:

1. *A sustained partial is booked as success.* The steady-state `RefreshPartiallySucceeded` branch in `_async_update_data` calls `_mark_success()` and returns the plant with the failed device's frozen registers riding along; `_mark_success()` resets `consecutive_failures`, so every tick counts as clean.
2. *Nothing escalates a sustained partial to recovery.* `_record_partial()` only throttles logging. `_schedule_reconnect` is set only on a topology *loss* (a device detected as gone), never on a device that stays present but whose reads keep failing — so a wedged bank never re-detects on its own.
3. *Battery sensors have no staleness gate* (the open #152 follow-up), so the frozen value is shown as current instead of going unavailable.

**Proposed fix (two composable parts)**

- *Recovery:* after N consecutive partials affecting the same device, escalate to a reconnect/re-detect (the path a manual reload already exercises) rather than serving frozen data forever.
- *Detection:* extend the staleness/availability gate to battery and AIO-module sensors per #152, so a frozen bank flips to unavailable.
- *Diagnostics:* retain the last partial-failure detail — which device dropped, with a timestamp — even after a subsequent clean poll. Today `last_partial_failures` is cleared on the next clean poll ([coordinator.py:218]), so the `Partial Refresh Failures` sensor's device-naming attributes (`last_failed_devices`, `last_failures`) only exist *while* a poll is actively failing. A user sees the cumulative count but cannot tell *what* dropped unless they catch it mid-stall — which makes intermittent failures effectively un-diagnosable after the fact.

**Confirmation still needed**

[...]

PR #183, opened 2026-06-17
fix: coordinator robustness + battery staleness gate (#176)

dewet22:
Re-scoped #176 after Nick's diagnostic logs (the original "sustained partial freezes data forever" hypothesis didn't hold — his partials are intermittent and the real frozen-battery path is the modbus library holding last-good on a sub-bus splice, which it reports as success). Three independent fixes, one atomic commit each:

**Discard a half-initialised client on a failed reconnect.** A reconnect whose `detect()` timed out left a connected client with no capabilities attached; the timeout-tolerance path kept it, the next tick skipped `_connect()` (the socket still reported connected), and `refresh()` then raised `PlantNotDetected`. It self-healed after a couple of noisy ERROR ticks. `_connect()` now drops the client on any failure before capabilities are established, so the next tick reconnects and re-detects cleanly. `PlantTopologyMismatch` is still handled in-line and unaffected.

**Battery staleness gate (#152 follow-up).** Battery sensors had no staleness gate, so a held/frozen bank showed a confidently-wrong value indefinitely — the frozen-SOC symptom behind the reports. They now go unavailable once the backing IR bank stops committing past the same ceiling the inverter sensors already use (max 5 min / 10× scan interval), mirroring the existing inverter gate. AIO per-module sensors aren't covered yet: `AioBatteryModule` has no register-getter, so resolving its registers needs a library-side change first, which I'll raise separately.

**Retain partial-failure diagnostics.** The "Partial Refresh Failures" sensor used to forget which device dropped the moment a poll recovered, making intermittent failures un-diagnosable after the fact. It now retains the last detail plus a `last_partial_at` timestamp, so a recovered flaky device stays identifiable.

All TDD — each fix has a regression test. Full suite (332) + mypy + ruff green locally.

Re #176, #152.

### DIFF
--- a/custom_components/givenergy_local/coordinator.py
+++ b/custom_components/givenergy_local/coordinator.py
@@ -441,40 +441,54 @@ class GivEnergyUpdateCoordinator(DataUpdateCoordinator[Plant]):
         — required for three-phase, AIO-HV, EMS and other non-default topologies.
         """
         self._client = Client(host=self.host, port=self.port)
-        await self._client.connect()
-        topology_confirmed = True
         try:
-            # The library's battery sweep probes additional packs (0x33+) with a
-            # stingy default budget (probe_timeout=0.5s, probe_retries=1) and
-            # BREAKS at the first non-responding slot — so a single transiently
-            # slow BMS during a reconnect truncates the whole battery chain, and
-            # that reduced topology then gets persisted (a 2nd battery silently
-            # vanished after a reconnect this way). Because of the break, a detect
-            # probes at most one empty slot, so a generous probe budget costs only
-            # ~one extra slow probe on a cold sweep — cheap insurance against
-            # dropping a real pack.
-            await self._client.detect(
-                prior=self._prior_capabilities,
-                probe_timeout=PROBE_TIMEOUT_SECONDS,
-                probe_retries=PROBE_RETRIES,
-            )
-        except PlantTopologyMismatch as exc:
-            topology_confirmed = await self._handle_topology_mismatch(exc)
-        if self._client is None:
-            # The entry was unloaded/reloaded while _handle_topology_mismatch
-            # yielded (retry sleeps) and async_close() discarded the client —
-            # nothing to confirm against, we're shutting down.
-            return
-        confirmed = self._client.plant.capabilities
-        if topology_confirmed and self._on_topology_healed is not None and confirmed is not None:
-            # Full expected topology is present — let the caller clear any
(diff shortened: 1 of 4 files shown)


custom_components/givenergy_local/coordinator.py
```diff
@@ -441,40 +448,54 @@ async def _connect(self) -> None:
...
+            ):
+                # Full expected topology is present — let the caller clear any
+                # stale "device missing" repair and reconcile entities against the
+                # confirmed topology (covers both restart and mid-session recovery).
+                await self._on_topology_healed(confirmed)
+            self._last_inverter_time = None
+            self._unchanged_ticks = 0
+            self._active_tick = 0
+            _LOGGER.info("Connected to inverter at %s:%s", self.host, self.port)
+        except Exception:
```
dewet22-codex: This cleanup is what prevents a connected-but-capability-less client from surviving a failed `_connect()`, but `asyncio.CancelledError` does not inherit from `Exception` on supported Python versions. If HA cancels this update while `connect()` or `detect()` is in flight, the half-initialized `self._client` can remain attached and the next tick can hit the same `PlantNotDetected` path this PR is trying to eliminate. Please make the cleanup run for cancellation as well, while still re-raising the cancellation, e.g. use a `BaseException` cleanup boundary here or an equivalent `try`/`finally` pattern.

dewet22-codex (changes requested):
Thanks, the reconnect, battery staleness, and partial-failure diagnostic changes are well targeted. I found one cleanup gap in the new reconnect protection that should be addressed before merge.

custom_components/givenergy_local/coordinator.py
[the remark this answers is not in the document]
dewet22: Good catch — fixed in b22ab3a: the cleanup boundary now catches `BaseException`, so a `CancelledError` during `_connect()` also discards the half-initialised client (re-raised unchanged). Added a regression test (`test_cancelled_connect_discards_client`).

dewet22: Good catch — fixed in b22ab3a: the cleanup boundary now catches `BaseException`, so a `CancelledError` during `_connect()` also discards the half-initialised client (re-raised unchanged). Added a regression test (`test_cancelled_connect_discards_client`).

dewet22-codex (approved):
Re-reviewed the cancellation cleanup fix. The `BaseException` boundary now covers `CancelledError` while re-raising it, and the regression test exercises the leaked-client case. No remaining actionable concerns from me.

PR #185, opened 2026-06-18
feat: AIO-module staleness gate + shared _StaleIRGate mixin (#152)

dewet22:
Final piece of the #152 staleness work, now unblocked by givenergy-modbus 2.5.0 (modbus#273). Per-module AIO battery sensors now go unavailable when their backing IR bank stops committing — the same gate the inverter and battery sensors already use — so a frozen/held module reads unavailable instead of a confidently-wrong value, consistent with the #176 fix for LV packs.

The gate logic, previously duplicated across the inverter and battery sensors, is pulled into a shared `_StaleIRGate` mixin (register resolution + ceiling + `register_age()` check) — the inverter/battery refactor is behaviour-preserving.

HV `Bmu` per-module gating stays out of scope: those registers are stride-offset per module and need a different approach (gated on modbus#265, tied to #179).

Verified end-to-end against the real 2.5.0 `AioBatteryModule.REGISTER_GETTER` (`v_cell_01`→IR(60), `t_cell_24`→IR(113), `module_address`→() graceful). Full suite (335) + mypy + ruff green.

With this, the #152 staleness work and the re-scoped #176 are fully addressed — I'd propose closing both on merge.

Re #152, #176, #192.

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -1596,7 +1596,9 @@ class GivEnergyBatterySensor(
         return self.entity_description.value_fn(batteries[self._battery_index])
 
 
-class GivEnergyAioModuleSensor(CoordinatorEntity[GivEnergyUpdateCoordinator], SensorEntity):
+class GivEnergyAioModuleSensor(
+    _StaleIRGate, CoordinatorEntity[GivEnergyUpdateCoordinator], SensorEntity
+):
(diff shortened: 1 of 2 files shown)


dewet22-codex (approved):
Looks good. The shared stale-IR gate preserves the existing inverter/LV behavior while extending the same register-age check to AIO module sensors using the module device address, and the added tests cover the fresh/stale/never-committed and absent-module cases that matter for this change.

Issue #194, opened 2026-06-23
AIO/Gateway: first-class Gateway-connection support (#95 phase 2)


dewet22:
Follow-up to #95. Phase 1 (v1.3.5, #193) surfaced the HV battery stack and added the battery-data-only option for **per-AIO** connections. Phase 2 is supporting a connection pointed at the **Gateway** itself.

## Why
In a Gateway-fronted parallel-AIO setup the Gateway is the real controller. Connecting to it gives data the per-AIO connections can't:
- **`p_load` — true whole-house consumption.** This is the proper fix for the "House Consumption Today" oddity reported in #95 (the per-AIO derivation inflates overnight when the AIOs charge from grid).
- **Per-AIO aggregates** — `parallel_aio_num`/`online`, per-AIO SoC/power/energy, `p_aio_total`.

The modbus library already decodes all of this (`Plant.gateway`, `GatewayV1`/`V2`), so this is integration-side surfacing.

## Prerequisites / unknowns
- **Verify the integration can connect to a `Model.GATEWAY` device-type at all** — today `GATEWAY` is only handled for naming (`_MODEL_NAMES`/`_device_kind`); nothing in `coordinator.py`/`__init__.py` drives a Gateway as the connected device.
- The Gateway is **read-only** in the library (no Gateway write commands), so control stays per-AIO.
- Needs a wire **capture from a real Gen-1 Gateway** — @AberDino has offered to test.

## Related smaller items still outstanding from #95
- [ ] Surface the extra inverter read fields: `grid_port_max_power_output`, `inverter_max_power`, `model` (and `battery_max_power`, pending a modbus DTC-lookup entry).
- [ ] `skip_if_none` / suppress the inapplicable inverter sensors on AIO (MPPT Count, Solar Diverter, and the single-phase-only fields that read "Unknown").
- [ ] Defensive dedup-by-serial in the per-AIO-module loop (robustness; real serials are unique).

PR #195, opened 2026-06-23
feat: extra inverter power sensors + AIO sensor hygiene (#194)

dewet22:
## Summary
The three capture-independent items from #194 (follow-ups to #95).
- **New diagnostic sensors:** Grid Export Power Limit (`grid_port_max_power_output` — the G99 export cap AberDino asked for), Inverter Max Power, Battery Max Power. The two DTC-computed ones skip when their value is None (unmapped device-type code) rather than showing "Unknown". `model` not added — it's already in the device info.
- **AIO hygiene:** `skip_if_none` on the inverter sensors that read None on an AIO (MPPT Count, Inverter Export Total, Fault Messages, Inverter Errors, Battery Discharge This Year), so they drop instead of surfacing as permanently-Unknown orphans.
- **Dedup AIO modules by serial** — a duplicate/placeholder module serial no longer collides on unique_id (the error flood seen testing #95 against the mock).

Gateway-connection support stays tracked in #194.

## Test plan
- 456 tests pass (3 new); ruff + mypy clean.
- **Live check on a real single-phase inverter:** confirm the 3 new sensors populate (especially the G99 export limit) and that no previously-present inverter sensor vanished from the `skip_if_none` additions — `e_inverter_export_total` (extended HR block) is the one to watch.

## Release Notes

* **New Features**
  * Added three new inverter power-limit sensors: Grid Export Power Limit, Inverter Max Power, and Battery Max Power.

* **Bug Fixes**
  * Improved handling of inverter diagnostic sensors—they no longer appear as "Unknown" when unavailable on certain models.
  * Fixed duplicate battery module devices appearing during setup.

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -220,8 +220,10 @@ INVERTER_SENSORS: tuple[GivEnergyInverterSensorDescription, ...] = (
         value_fn=lambda inv: (
             ", ".join(inv.inverter_fault_messages) if inv.inverter_fault_messages else None
         ),
-        # None on AIO (no fault block) — drop rather than show "Unknown" (#194).
-        skip_if_none=True,
+        # Deliberately NOT skip_if_none: value_fn returns None for the healthy
+        # *empty fault list* too, not just the unsupported-on-AIO case. Skipping
+        # would drop this on a healthy single-phase install at setup, leaving a
+        # later fault nowhere to surface until a reload (Codex review, #194).
         entity_category=EntityCategory.DIAGNOSTIC,
     ),
     # ChargeStatus mapping shipped in modbus 2.3.1 (#222) — render the
@@ -1410,8 +1412,15 @@ async def async_setup_entry(
         # Skip a module whose serial we've already created entities for: a
         # duplicate serial (placeholder/garbled, or a repeated BMS read) would
         # otherwise collide on unique_id and HA would drop the second module's
-        # entities with an error per cell (#194).
+        # entities with an error per cell (#194). Log it loudly — real modules
+        # have unique serials, so a duplicate means a module is missing from HA.
         if module.serial_number in seen_module_serials:
+            _LOGGER.warning(
+                "Skipping AIO battery module at index %d: duplicate serial %s "
+                "(real modules have unique serials — check for a placeholder/garbled read)",
+                module_index,
+                module.serial_number,
+            )
             continue
         seen_module_serials.add(module.serial_number)
         entities.extend(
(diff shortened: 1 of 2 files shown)


dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -218,6 +220,8 @@ def _grid_import_power(inv: InverterModel) -> float | None:
         value_fn=lambda inv: (
             ", ".join(inv.inverter_fault_messages) if inv.inverter_fault_messages else None
         ),
+        # None on AIO (no fault block) — drop rather than show "Unknown" (#194).
+        skip_if_none=True,
```
dewet22-codex: I think this one is different from the other AIO-only `None` cases. The `value_fn` above also returns `None` when `inv.inverter_fault_messages` is an empty list, which is the normal healthy state on a supported inverter. Because `skip_if_none` is evaluated only during platform setup, a healthy install would never create the Fault Messages entity, so a later fault message has nowhere to surface until the entry is reloaded. I would leave this descriptor unskipped, or change the value function to distinguish an unsupported `None` block from an empty-but-supported fault list.

custom_components/givenergy_local/sensor.py
```diff
@@ -1360,9 +1403,17 @@ async def async_setup_entry(
...
+    seen_module_serials: set[str] = set()
     for module_index, module in enumerate(coordinator.data.aio_battery_modules):
         if not module.is_valid():
             continue
+        # Skip a module whose serial we've already created entities for: a
+        # duplicate serial (placeholder/garbled, or a repeated BMS read) would
+        # otherwise collide on unique_id and HA would drop the second module's
+        # entities with an error per cell (#194).
+        if module.serial_number in seen_module_serials:
+            continue
```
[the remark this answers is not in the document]
dewet22: Added the warning log in f5c5c23 — agree, a module dropped for a duplicate/placeholder serial should be diagnosable rather than silently missing.

custom_components/givenergy_local/sensor.py
dewet22: Good catch — reverted `skip_if_none` on Fault Messages in f5c5c23. Its `value_fn` returns None for the healthy empty-fault list too (not just the unsupported-AIO case), so skipping would drop it on a healthy single-phase install and a later fault couldn't surface until reload. The other four additions stay — they're non-None when healthy (e.g. `inverter_errors` reads raw `0`).

dewet22-codex:

dewet22-codex (approved):
I’ve re-reviewed the latest head and have no concerns to add. The Fault Messages descriptor is no longer skipped at setup, the regression assertion now captures that case, duplicate AIO module serials are still deduplicated, and the skip now emits a warning so the missing module is diagnosable. CI is green.

PR #202, opened 2026-06-23
feat: surface cold-start baseline holds + bump modbus floor to 2.5.5

dewet22:
## Summary

modbus 2.5.5 ([#289](https://github.com/dewet22/givenergy-modbus/issues/289)) adds cold-start battery-baseline confirmation — a battery bank is held one extra poll until a second read corroborates it, so a transient first-frame sub-bus splice can't poison the baseline. It exposes a per-device `cold_start_held_count`.

This:
- **Bumps the modbus floor to `>=2.5.5`** (manifest + pyproject + uv.lock) to pick up #289 — the cold-start half of the splice-guard work, completing the battery-offline-robustness saga (after 2.5.2's poisoned-baseline self-heal and 2.5.4's heal-after-insistence).
- **Surfaces `cold_start_held_count` as a fifth coordinator diagnostic** ("Comms Cold Start Holds"), reusing the existing reset-aware comms-counter accumulation and the per-device breakdown attribute. It's a benign "establishing baseline" signal — deliberately distinct from "Comms Splice Guard Holds" (which means corruption is being held).
- **Groups the five noise-floor diagnostics under a "Comms" name prefix** (CRC Errors, Splice Guard Rejections/Holds, Read Retries, Cold Start Holds) so they cluster together in the device's Diagnostic section. Display-name only — each sensor's `key`/`unique_id` is unchanged, so the four already shipped in v1.3.8 keep their entity_id and history; only the friendly name moves.

No other behaviour change: #289's "battery telemetry appears ~1 poll later at cold start" is already tolerated (absent battery registers render unavailable, well under the 300s staleness ceiling).

## Test plan
- [x] `uv run pytest -q` — 470 passed
- [x] Verified `Plant.cold_start_held_count` is present in modbus 2.5.5
- Extended the coordinator accumulation tests and the sensor descriptor/default-zero tests to cover the fifth counter.

### DIFF
--- a/custom_components/givenergy_local/coordinator.py
+++ b/custom_components/givenergy_local/coordinator.py
@@ -192,6 +192,9 @@ class GivEnergyUpdateCoordinator(DataUpdateCoordinator[Plant]):
         self.splice_rejections_by_device: dict[int, int] = {}
         self.splice_holds_by_device: dict[int, int] = {}
         self.read_retries_by_device: dict[int, int] = {}
+        # Cold-start battery-baseline holds (modbus #289, 2.5.5): a benign
+        # "establishing baseline" signal, distinct from splice_holds (corruption).
+        self.cold_start_held_by_device: dict[int, int] = {}
         # Last cumulative value seen per library attr per device, for delta-ing.
         self._comms_last_seen: dict[str, dict[int, int]] = {}
         # Length of the current unbroken run of partial polls, used only to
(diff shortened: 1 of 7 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the dependency and diagnostic handoff and have no blocking concerns. The new modbus 2.5.5 counter is folded through the existing reset-aware per-device accumulator, exposed with the same total/breakdown conventions as the other comms diagnostics, and the existing sensors retain their stable unique-id keys despite the friendly-name grouping. Manifest, project metadata and lockfile agree on the dependency floor. The full local suite passes (470 tests).

PR #204, opened 2026-06-23
feat: AIO feedback follow-ups — drop bogus HV Battery Power, gate PV sensors, add-time battery-data-only (#95)

dewet22:
## Summary

Addresses AberDino's v1.3.8 testing round on [#95](https://github.com/dewet22/givenergy-hass/issues/95) (Gen-1 Gateway + 2 AIOs):

- **Drop the HV-stack "Battery Power" sensor.** On a real AIO BCU it reads a 0xFFFF sentinel (~65.53 kW when discharging, 0 when idle) — never a usable figure. The inverter-level Battery Power is authoritative. (His explicit ask.)
- **Gate three inverter sensors on AIO topology, not None.** MPPT Count, Solar Diverter Energy and Battery Discharge This Year read 0 / 0.0 / a static figure on an AIO — *not* None — so the v1.3.6 `skip_if_none` never dropped them. New `skip_if_aio` flag, evaluated against the plant carrying AIO battery modules.
- **Offer battery-data-only at add time.** Previously options-only, so adding a parallel-mode AIO created all the inverter sensors first and they then went unavailable. The toggle is now on the initial config-flow step too, landing in `entry.options` (read everywhere).

Tracked separately, not in this PR: the "PV Generation" naming on AIO (a register-identity question for givenergy-modbus) and first-class Gateway connection ([#194](https://github.com/dewet22/givenergy-hass/issues/194)).

## Test plan
- [x] `uv run pytest -q` — 472 passed (+2: AIO sensor gating, add-time battery-data-only)
- Real-hardware confirmation to come from AberDino on the next release.

## Summary by CodeRabbit

* **New Features**
  * Added "Battery data only" configuration option during initial setup, allowing users to suppress controls and non-essential system sensors.

* **Bug Fixes**
  * Removed inverter sensors that are not applicable to battery-only systems, reducing unnecessary data display.

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -1512,9 +1512,12 @@ async def async_setup_entry(
     inverter = coordinator.data.inverter
     capabilities = coordinator.data.capabilities
     is_three_phase = bool(capabilities and capabilities.is_three_phase)
-    # An AIO (battery-only) plant carries per-module battery devices; its inverter
-    # exposes PV/solar fields that read meaningless values, so gate those out (#95).
-    is_aio = bool(coordinator.data.aio_battery_modules)
+    # A pure All-in-One is battery-only, so its inverter's PV/solar fields read
+    # meaningless values — gate those sensors out (#95). Key off the detected model
+    # (restored from the on-disk capabilities cache), NOT decoded module telemetry:
+    # an AIO whose modules drop out on the setup poll must still be recognised, and
+    # ALL_IN_ONE_HYBRID genuinely has PV so is deliberately excluded.
+    is_aio = bool(capabilities) and capabilities.device_type is Model.ALL_IN_ONE
     battery_data_only = entry.options.get(CONF_BATTERY_DATA_ONLY, DEFAULT_BATTERY_DATA_ONLY)
 
     entities: list[SensorEntity] = []
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -56,6 +56,11 @@ class GivEnergyInverterSensorDescription(SensorEntityDescription):
     value_fn: Callable[[InverterModel], Any] = field(default=lambda _: None)
     # If True, the entity is not created when value_fn returns None at first refresh.
     skip_if_none: bool = False
+    # If True, the entity is not created on an AIO (battery-only) plant, where the
+    # field reads a meaningless value rather than None (MPPT count 0, a solar
+    # diverter total, a static yearly discharge figure), so skip_if_none can't
+    # catch it (#95).
+    skip_if_aio: bool = False
     # If True, the entity is not created on three-phase inverters, where the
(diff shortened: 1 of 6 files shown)


dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -1506,6 +1512,9 @@ async def async_setup_entry(
     inverter = coordinator.data.inverter
     capabilities = coordinator.data.capabilities
     is_three_phase = bool(capabilities and capabilities.is_three_phase)
+    # An AIO (battery-only) plant carries per-module battery devices; its inverter
+    # exposes PV/solar fields that read meaningless values, so gate those out (#95).
+    is_aio = bool(coordinator.data.aio_battery_modules)
```
dewet22-codex: Red (blocking): this ties the fixed entity set to whether module telemetry decoded on the setup poll, rather than to the plant model. A detected `Model.ALL_IN_ONE` can have `aio_battery_modules == []` when its module caches are absent or temporarily fail to decode; the three meaningless inverter sensors are then created and remain in the registry even after module data recovers. I think the stable gate here is `capabilities.device_type is Model.ALL_IN_ONE` (which also avoids suppressing these fields on `ALL_IN_ONE_HYBRID`), with a regression where the model is AIO but the initial module list is empty.

dewet22-codex:

dewet22-codex (approved):
I’ve re-reviewed the latest head and the blocking topology issue is addressed. The three pure-AIO-only suppressions now key off the stable detected `Model.ALL_IN_ONE` identity rather than module telemetry, with regression coverage for an AIO whose initial module list is empty; `ALL_IN_ONE_HYBRID` remains unaffected. The focused config/sensor suite passes locally (119 tests) and the full suite passes (473 tests).

custom_components/givenergy_local/sensor.py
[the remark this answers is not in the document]
dewet22: Good catch — switched the gate to `capabilities.device_type is Model.ALL_IN_ONE`, keyed off the cached model rather than the decoded module list, so a transient cold-start module drop can't orphan these sensors. I used the model identity check rather than `aio_battery_module_addresses` so `ALL_IN_ONE_HYBRID` (which has genuine PV) is left alone.

dewet22: Agreed — moved the gate to `capabilities.device_type is Model.ALL_IN_ONE`, which is stable across a cold-start module-decode failure and also excludes `ALL_IN_ONE_HYBRID` (genuine PV). Added a regression where the model is AIO but the initial module list is empty (`test_aio_model_suppresses_pv_sensors_even_with_empty_modules`).

dewet22-codex:
I’m starting another re-review of the latest changes now.

dewet22-codex (approved):
I’ve rechecked the unchanged approved head after the review-thread replies. Both topology threads are now resolved, the AIO model gate and empty-module regression remain correct, and the focused config/sensor suite passes again locally (119 tests). My approval stands.

PR #205, opened 2026-06-23
feat: trim EMS controller device to its real role (#201)

dewet22:
Closes #201.

## Summary

On an EMS plant the device at modbus `0x11` is an EMS **controller**, not an inverter, but the integration gave it the full inverter entity set — ~120 mostly-unavailable sensors plus redundant inverter controls. Validated against @njhcarroll's real EMS plant (#52): his controller device carried **128 entities**, the bulk unavailable.

This gates on `coordinator.data.ems is not None` (the established EMS signal, mirroring the existing `time.py` EMS/Smart-Load branch):

- **Sensors:** suppress the whole inverter sensor loop on EMS and surface the `Ems` plant-level aggregates instead — **EMS Status**, **EMS Managed Inverter Count**, **EMS Calculated/Measured Load Power**, **EMS Grid Meter Power**, **EMS Total Battery Power**, **EMS Remaining Battery Energy**. Names are prefixed `EMS` so they group on the controller device.
- **Controls:** suppress the inverter-level controls on EMS across switch / number / select / time. The already-gated EMS slot/target controls (and Flexi EMS Control) are authoritative and remain.

A directly-connected inverter entry is unaffected (`ems is None` → it keeps its full sensor + control set). The managed-inverter rollup devices (#199) are unchanged. The battery out-of-spec binary sensor self-gates on `batteries` and stays (it's battery data).

Net effect on Nick's controller: ~128 entities → a lean set (7 `EMS …` sensors + the coordinator/Comms diagnostics + the EMS slot controls).

## Scope notes
- The #199 managed-inverter serial-collision with directly-connected inverters is tracked separately in #203.
- `_device_kind`/`_MODEL_NAMES` lacking an `EMS_COMMERCIAL` label is pre-existing and left as-is.

## Test plan
- [x] `uv run pytest -q` — 474 passed (4 new EMS tests: inverter sensors + controls suppressed on EMS; EMS aggregates created/read; absent on non-EMS)
- Live validation against an EMS mock / Nick's plant to follow before release.

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -9,6 +9,7 @@ from typing import Any
 from givenergy_modbus.model.aio_battery import AioBatteryModule
 from givenergy_modbus.model.battery import Battery, BatteryMaintenance
 from givenergy_modbus.model.devices import InverterSummary
+from givenergy_modbus.model.ems import Ems
 from givenergy_modbus.model.hv_bcu import Bcu, HvStack
 from givenergy_modbus.model.inverter import (
     BatteryCalibrationStage,
@@ -125,6 +126,11 @@ def _bcu_attr(name: str) -> Callable[[Bcu], Any]:
 @dataclass(frozen=True, kw_only=True)
 class GivEnergyManagedInverterSensorDescription(SensorEntityDescription):
     value_fn: Callable[[InverterSummary], Any] = field(default=lambda _: None)
+
+
+@dataclass(frozen=True, kw_only=True)
+class GivEnergyEmsSensorDescription(SensorEntityDescription):
+    value_fn: Callable[[Ems], Any] = field(default=lambda _: None)
 
 
 def _summary_attr(name: str) -> Callable[[InverterSummary], Any]:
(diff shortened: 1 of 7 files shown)


dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -1507,20 +1596,32 @@ async def async_setup_entry(
...
+    # registers (PV/battery/grid/AC) are absent, so the whole inverter sensor set
+    # would surface as permanently-unavailable orphans (#201). Suppress them and
+    # emit the EMS plant-level aggregates instead.
+    #
     # Battery-data-only (#95): a parallel-mode AIO is controlled by its Gateway,
     # so its own inverter-level sensors (PV/grid/load/derived consumption) are
     # misleading. Drop them; keep the battery pack / HV stack / module / coordinator
     # data below. Control platforms suppress themselves the same way.
-    if not battery_data_only:
+    if not battery_data_only and ems is None:
```
dewet22-codex, line 1624: Red (blocking): suppressing creation only affects fresh installs. On an upgraded EMS entry, Home Assistant retains the existing entity-registry rows when a platform no longer adds those entities, so Nick’s controller will still carry the old inverter sensors and controls as orphaned/unavailable entries rather than falling from ~128 to the lean set described here. I think the EMS setup path needs an explicit, narrowly-scoped registry reconciliation for the retired inverter sensor/control unique IDs (while preserving coordinator, battery, managed-inverter and EMS-specific entities), with an upgrade regression that seeds an old entity and confirms it is removed.

dewet22: Fixed in af0851a — you're right that suppressing creation alone left orphaned rows on an upgraded EMS entry. Added `_reconcile_ems_entities` in `__init__.py`: on EMS setup (before the platforms load) it removes registry rows whose unique_id matches the controller serial + a retired inverter sensor/control key, built from the exact description tuples the EMS path now skips — preserving coordinator, battery, AIO, managed-inverter and EMS-specific entities. Added an upgrade regression that seeds the old inverter rows across the sensor + all four control platforms and asserts they're removed, while a coordinator diagnostic and an EMS aggregate sensor survive.

dewet22-codex:

dewet22-codex (approved):
I’ve re-reviewed the latest head and the upgrade path now addresses my earlier concern. The reconciliation is scoped to exact controller-serial and retired-description IDs, runs before platform setup, and preserves the EMS, coordinator, battery and managed-inverter entities. The focused EMS tests and full suite (475 tests) pass locally. I have no blocking concerns remaining.

dewet22-codex:
I’m starting another re-review on the rebased head now.

dewet22-codex (approved):
I’ve re-reviewed the rebased head. The upgrade reconciliation commit is patch-identical to the version previously reviewed; the only surrounding change brought in from main is the new AIO sensor predicate, and EMS entries bypass that inverter-sensor loop. The focused EMS tests pass, as does the full suite (478 tests). I have no blocking concerns.

Issue #188, opened 2026-06-22
Not able to set value for Battery AC Charge Limit (which is mistitled)


njhcarroll:
HI Dewet

I'm ploughing through preparing the setup yaml for predbat to use GE Local.
I'll summarise some helpers as feature requests for the integration that you might want to consider later.
Predbat needs to be able to change the charge/discharge power either by setting the value or by changing the percentage. I discovered by changing "inverter charge power percentage" and "inverter discharge power percentage" on portal that they map to "Battery AC charge limit" and "Battery AC discharge limit" in the configuration section on GE Local.  If I change a value on the portal it immediately is reflected on GE local.  However, if I try to change the value on GE local, I get an error "failed to perform the action set value HR(313) is not permitted for AC inverter"

Looks like I can't try the integration until it will allow me to write.

Looking at the logs - I can see the following error is generated each time I try to change the value in GE Local

Log details (ERROR)
Logger: homeassistant.components.websocket_api.http.connection
Source: components/websocket_api/commands.py:281
Integration: Home Assistant WebSocket API (documentation, issues)
First occurred: 16:52:34 (6 occurrences)
Last logged: 17:14:36

[546331968832] Unexpected exception
[546364397888] Unexpected exception
[545579703552] Unexpected exception
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/components/websocket_api/commands.py", line 281, in handle_call_service
    response = await hass.services.async_call(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<7 lines>...
    )
    ^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2827, in async_call
    response_data = await coro
                    ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2870, in _execute_service
    return await target(service_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 949, in entity_service_call
    response_data = await _async_handle_entity_calls(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<5 lines>...
    )
    ^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 828, in _async_handle_entity_calls
    single_result = await entity.async_request_call(_with_context(entity, coro))
[...]

Issue #175, opened 2026-06-16
Auto generated dashboard not registering both inverters in EMS setup


njhcarroll:
I'm impressed with the autogenerated dashboard functionality.
For the EMS system, it only seems to recognise the first of my inverters attached to the EMS.  Also the PV values are all zero.

<img width="1206" height="2622" alt="Image" src="https://github.com/user-attachments/assets/3f02c1a7-e285-4900-94fe-9d608eb5634b" />
<img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/407e2bc6-48f0-4367-9461-d52245a2c0f1" />

dewet22:
Thanks, Nick — a good catch on top of everything you've already helped shake out on the EMS side. The screenshots made it clear.

What's happening: the auto-generated dashboard is built around a *single* inverter, and on an EMS system it lands on one of the individual inverters rather than the EMS controller. The controller is the device that carries the aggregated figures — your EMS device has the real PV total (~9.8 kWh today), while the individual inverter the dashboard picked reports no PV of its own, which is why the solar panels read zero. So both symptoms (only one inverter, and PV at zero) come from the same root: the dashboard doesn't yet understand the EMS "controller plus several inverters" topology.

That's a genuine gap rather than a quick toggle — making the dashboard EMS-aware means enumerating the inverters behind the controller and aggregating across them. Given how far the rest of EMS support has come, it feels like the natural next piece, and I'll use this issue to track it, with your screenshots as the reference for what it should show.

In the meantime all the underlying entities are present and updating — it's only the auto-dashboard's single-inverter assumption that misrepresents them. A manual dashboard can point cards straight at the EMS controller's aggregated sensors (PV Energy, Grid Power, and the like); anything the controller doesn't already aggregate is exactly what this dashboard work will pull together.

dewet22:
Hi Nick — the EMS dashboard issue is fixed in **v1.3.16** (just out on HACS).

[...]

PR #206, opened 2026-06-24
fix: keep meaningful inverter sensors on EMS controllers (#201 narrowing)

dewet22:
Narrows #205 (shipped in v1.3.11), which over-suppressed.

## The regression
#205 dropped the **entire** inverter sensor set on EMS controllers, on the premise they were "mostly unavailable." That premise was a **CRC artifact** — the mock's `0x11` frames CRC-fail and tester plants are comms-noisy, so the sensors *read* as unavailable but aren't inapplicable. Loading the committed `ems_2_inv_3_bat_a/ems_arm1036_30min.log` capture, the inverter facade over the `0x11` cache carries real plant data:

`e_pv_generation_today=14.1`, `e_pv_generation_total=24172.3`, `p_grid_out=94`, grid totals 12122/28618, `battery_soc=45`, `v_battery=52.68`, `v_ac1=242.7`, `f_ac1=49.9`, `t_inverter_heatsink=45.6`, `status=NORMAL`.

Corroborated by dewet22's #175 comment ("the EMS device has the real PV total") and @njhcarroll's #52 ("inverter SOC readings broadly match the portal"). So v1.3.11 removed genuinely useful data.

## The narrowing
- **Keep the inverter sensors on EMS.** The inverter loop runs regardless of EMS again.
- **Gate only the broken derived sensor** — House Consumption Today (`e_consumption_today`) computes a wrong per-controller figure on an EMS (e.g. −9.8), so it gets a new per-description `skip_if_ems` flag (same shape as `skip_if_aio`). `Ems.e_active_generation_total` (583) ≠ the PV total, so it can't substitute — the PV figures must come from the inverter sensors.
- **Keep the inverter-control suppression** (EMS slots are authoritative; #188 shows the inverter-level writes are rejected anyway).
- **Drop the duplicate `ems_status`** EMS aggregate — the restored inverter "Status" sensor covers it.
- **Narrow `_reconcile_ems_entities`** to remove only the controls + House Consumption + `ems_status` on upgrade, retaining (and re-creating) the inverter sensors.

## Test plan
- [x] `uv run pytest -q` — 479 passed (EMS tests rewritten: inverter sensors kept on EMS except House Consumption; upgrade reconciliation narrowed; `skip_if_ems` flag unit test)
- [x] Verified against the EMS capture: kept sensors read real values; `e_consumption_today` is the −9.8 being gated
- Live confirmation from Nick on #52 to follow once v1.3.12 ships.

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined how inverter-level entities are managed when the system is part of an EMS plant installation.
  * Inverter sensors are now handled with finer granularity—only specific derived sensors are suppressed rather than all inverter sensors.
  * Fixed incorrect sensor readings on EMS controllers by properly suppressing the "House Consumption Today" sensor on EMS plant installations.

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -62,6 +62,11 @@ class GivEnergyInverterSensorDescription(SensorEntityDescription):
     # diverter total, a static yearly discharge figure), so skip_if_none can't
     # catch it (#95).
     skip_if_aio: bool = False
+    # If True, the entity is not created on an EMS plant. The 0x11 controller's
+    # direct registers are meaningful (PV/grid/battery/AC), but a few *derived*
+    # sensors (House Consumption = inverter-battery-grid) are computed for a single
+    # unit and read wrong on a controller, so gate just those (#201).
+    skip_if_ems: bool = False
     # If True, the entity is not created on three-phase inverters, where the
     # underlying field is single-phase-only (e.g. a p_pv1+p_pv2 sum) and so would
     # be meaningless or surface as a permanently-unavailable orphan entity.
@@ -658,6 +663,10 @@ INVERTER_SENSORS: tuple[GivEnergyInverterSensorDescription, ...] = (
             inv, "e_consumption_today", getattr(inv, "e_load_today", None)
         ),
         skip_if_none=True,
+        # Derived (inverter-battery-grid) — wrong on an EMS controller, where it
+        # computes a per-controller figure (e.g. -9.8) rather than whole-plant load;
+        # the EMS load aggregates cover it there instead (#201).
+        skip_if_ems=True,
         # Resolves per-model: e_load_today isn't in the single-phase LUT (the
         # derived value stays untracked for staleness), but on three-phase it
         # names the native register the value_fn falls back to (#152/#158).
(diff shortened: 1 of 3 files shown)


dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -658,6 +663,10 @@ def _grid_import_power(inv: InverterModel) -> float | None:
             inv, "e_consumption_today", getattr(inv, "e_load_today", None)
         ),
         skip_if_none=True,
+        # Derived (inverter-battery-grid) — wrong on an EMS controller, where it
+        # computes a per-controller figure (e.g. -9.8) rather than whole-plant load;
+        # the EMS load aggregates cover it there instead (#201).
+        skip_if_ems=True,
```
dewet22-codex, line 674: Red (blocking): the same EMS capture used to justify restoring the inverter sensors shows another load value that appears to need this gate. Its final cache has `p_load_demand=48 W`, while the EMS block reports `calc_load_power=489 W` (`measured_load_power=0`). The modbus model documents IR42 as the inverter busbar load, so on the EMS facade this appears controller-local in the same way this daily derivation is, rather than the whole-plant load. This is particularly visible because the bundled dashboard and recommended-entity set select `p_load_demand`, while the PR also exposes the EMS load aggregates. I think `p_load_demand` belongs under `skip_if_ems` as well unless there is field evidence that the 48 W reading is the intended plant load; a capture-backed assertion against the EMS aggregate would pin that boundary.

dewet22: Agreed and fixed in 6bb5802. Confirmed against the capture: `p_load_demand`=48 W vs `calc_load_power`=489 W (`measured_load_power`=0), so on the EMS facade it's the controller-local busbar load, not the plant load — added it to `skip_if_ems` (the reconciliation picks it up automatically, being data-driven off the flag). While there I checked the other two power flows: grid is fine (`p_grid_out`=94 vs `grid_meter_power`=−94 — same magnitude, opposite sign convention, so it *is* the plant grid) and battery power agrees (both 0), so those stay. A `{e_consumption_today, p_load_demand}` set assertion now pins the gated boundary against silent drift.

dewet22-codex:

dewet22-codex (approved):
I’ve re-reviewed the latest head and the blocking load-boundary concern is resolved. `p_load_demand` now shares the EMS gate with the derived daily consumption sensor, and the data-driven reconciliation includes both while retaining the validated inverter telemetry. The focused EMS tests and full suite (479 tests) pass locally. The bundled EMS dashboard still follows its existing controls-and-diagnostics-only path, so surfacing the restored telemetry there remains a separate non-blocking follow-up.

PR #208, opened 2026-06-24
feat: gate firmware-variable controls on register readability (#207)

dewet22:
Closes #207. The readability rubric from the #188 discussion: a control should only appear if its register actually reads on this device/firmware.

## Problem
Control entities (number/select/time) were created from capability gates alone, so a control could appear for a register the device/firmware doesn't carry — a non-functional slider, or one whose write the library rejects (cf. #188's HR313/314 on older firmware, and battery pause mode on systems that lack it).

## Change
Add a `skip_if_none` flag to the control description dataclasses — the control-side of the sensor `skip_if_none`. A flagged control is dropped when its value accessor (`value_fn`/`current_option_fn`/`slot_fn`) reads `None` at setup, guarded so a raising accessor (e.g. a field renamed in givenergy-modbus) skips just that control rather than aborting the platform (the same failure mode `_include_inverter_sensor` guards for sensors).

**Opt-in** to the genuinely firmware/device-variable controls:
- `battery_charge_limit_ac` / `battery_discharge_limit_ac` (HR313/314) — absent on older firmware (writability tracked in dewet22/givenergy-modbus#295).
- `battery_pause_mode` + `battery_pause_slot_start`/`_end` — the library notes pause is firmware-gated.

Core controls (charge slots, enable-charge, etc.) are deliberately **not** flagged — they should never vanish on a transient setup read-miss. The flag is there to extend to any control later.

## Test plan
- [x] `uv run pytest -q` — 482 passed (per-platform absent-when-`None` tests + the existing present-case coverage)
- The default fixtures set these registers to valid values, so existing control tests are unaffected; the gate only fires when a register reads `None`.

### DIFF
--- a/custom_components/givenergy_local/__init__.py
+++ b/custom_components/givenergy_local/__init__.py
@@ -478,6 +478,48 @@ def _reconcile_ems_entities(hass: HomeAssistant, entry: ConfigEntry, serial: str
             registry.async_remove(ent.entity_id)
 
 
+def _remove_stale_control(registry: er.EntityRegistry, serial: str, domain: str, key: str) -> None:
+    """Remove a readability-gated control's stale registry row, if present (#207)."""
+    entity_id = registry.async_get_entity_id(domain, DOMAIN, f"{serial}_{key}")
+    if entity_id is not None:
+        _LOGGER.info(
+            "Removing control %s: register absent on this device/firmware (#207)", entity_id
+        )
+        registry.async_remove(entity_id)
+
+
+def _reconcile_readability_gated_controls(
+    hass: HomeAssistant, coordinator: GivEnergyUpdateCoordinator
+) -> None:
+    """Remove control rows whose register is absent on this device/firmware (#207).
+
+    The control platforms skip creating a skip_if_none control when its register
+    reads None, but on an upgraded entry HA keeps the pre-existing row — an
+    orphaned, unavailable control the readability gate is meant to remove. Mirror
+    the EMS reconciliation: remove exactly those rows pre-platform, reusing each
+    platform's own gate so the readability logic isn't duplicated here.
+    """
+    # Local imports: the platforms import from this package, so importing them at
+    # module scope risks a load-order cycle. The gate helpers are the single source
+    # of truth for "is this control's register present".
+    from .number import AC_COUPLED_NUMBER_DESCRIPTIONS, _include_number
+    from .select import SELECT_DESCRIPTIONS, _include_select
+    from .time import TIME_DESCRIPTIONS, _include_time
+
+    inverter = coordinator.data.inverter
+    serial = coordinator.data.inverter_serial_number
+    registry = er.async_get(hass)
+    for number_desc in AC_COUPLED_NUMBER_DESCRIPTIONS:
+        if not _include_number(number_desc, inverter):
+            _remove_stale_control(registry, serial, "number", number_desc.key)
+    for select_desc in SELECT_DESCRIPTIONS:
+        if not _include_select(select_desc, inverter):
+            _remove_stale_control(registry, serial, "select", select_desc.key)
+    for time_desc in TIME_DESCRIPTIONS:
+        if not _include_time(time_desc, inverter):
+            _remove_stale_control(registry, serial, "time", time_desc.key)
+
+
 async def _async_reload_entry(hass: HomeAssistant, entry: ConfigEntry) -> None:
     """Reload the entry when its options change (registered as an update listener)."""
     await hass.config_entries.async_reload(entry.entry_id)
@@ -618,6 +660,10 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     # keeping them as orphaned/unavailable rows. Fresh installs match nothing.
     if coordinator.data.ems is not None:
         _reconcile_ems_entities(hass, entry, coordinator.data.inverter_serial_number)
+
+    # Remove control rows whose register is absent on this device/firmware (#207):
+    # the platforms skip creating them, but an upgraded entry keeps the stale row.
+    _reconcile_readability_gated_controls(hass, coordinator)
 
     await hass.config_entries.async_forward_entry_setups(entry, PLATFORMS)
 
--- a/custom_components/givenergy_local/__init__.py
+++ b/custom_components/givenergy_local/__init__.py
@@ -506,6 +506,13 @@ def _reconcile_readability_gated_controls(
     from .select import SELECT_DESCRIPTIONS, _include_select
     from .time import TIME_DESCRIPTIONS, _include_time
 
+    # A partial seed poll serves last-good with last_partial_failures set, so a None
+    # read may be a transient bank failure rather than structural absence. Removing
+    # rows now would lose history/customisation and the controls until a reload —
+    # reconcile only on a clean seed (#208 review).
+    if coordinator.last_partial_failures:
+        return
+
     inverter = coordinator.data.inverter
     serial = coordinator.data.inverter_serial_number
     registry = er.async_get(hass)
--- a/custom_components/givenergy_local/number.py
+++ b/custom_components/givenergy_local/number.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 from collections.abc import Callable
 from dataclasses import dataclass, field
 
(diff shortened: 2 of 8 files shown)


custom_components/givenergy_local/number.py
```diff
@@ -196,6 +205,23 @@ def _ems_number_descriptions() -> tuple[GivEnergyEmsNumberEntityDescription, ...
...
+    raises (a field renamed in givenergy-modbus) skips just that control, not the
+    whole platform.
+    """
+    if not description.skip_if_none:
+        return True
+    try:
+        return description.value_fn(inverter) is not None
+    except Exception:  # noqa: BLE001
+        _LOGGER.warning("Skipping number %s: value_fn raised at setup", description.key)
+        return False
```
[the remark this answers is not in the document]
dewet22: Added a skip log in 30cd7de, at **DEBUG** rather than WARNING: a None read here is the readability gate working as designed (the register is absent on this device/firmware, and it'd recur on every restart), not the duplicate/garbled anomaly the warn-on-skip convention targets — so WARNING would be the wrong severity and noisy. WARNING is kept for the genuine raise case (e.g. a renamed library field).

custom_components/givenergy_local/select.py
```diff
@@ -110,6 +118,22 @@ def _export_priority_current_option(inv: InverterModel) -> str | None:
...
+    device/firmware (e.g. battery pause mode on firmware without it). Guarded so a
+    raising accessor skips just that control, not the whole platform.
+    """
+    if not description.skip_if_none:
+        return True
+    try:
+        return description.current_option_fn(inverter) is not None
+    except Exception:  # noqa: BLE001
+        _LOGGER.warning("Skipping select %s: current_option_fn raised at setup", description.key)
+        return False
```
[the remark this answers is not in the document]
dewet22: Added a skip log in 30cd7de, at **DEBUG** rather than WARNING: a None read here is the readability gate working as designed (the register is absent on this device/firmware, and it'd recur on every restart), not the duplicate/garbled anomaly the warn-on-skip convention targets — so WARNING would be the wrong severity and noisy. WARNING is kept for the genuine raise case (e.g. a renamed library field).

custom_components/givenergy_local/time.py
```diff
@@ -201,6 +210,22 @@ def _ems_time_descriptions() -> tuple[GivEnergyEmsTimeEntityDescription, ...]:
...
+    device/firmware (e.g. the battery pause slot on firmware without it). Guarded so
+    a raising accessor skips just that control, not the whole platform.
+    """
+    if not description.skip_if_none:
+        return True
+    try:
+        return description.slot_fn(inverter) is not None
+    except Exception:  # noqa: BLE001
+        _LOGGER.warning("Skipping time %s: slot_fn raised at setup", description.key)
+        return False
```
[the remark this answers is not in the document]
dewet22: Added a skip log in 30cd7de, at **DEBUG** rather than WARNING: a None read here is the readability gate working as designed (the register is absent on this device/firmware, and it'd recur on every restart), not the duplicate/garbled anomaly the warn-on-skip convention targets — so WARNING would be the wrong severity and noisy. WARNING is kept for the genuine raise case (e.g. a renamed library field).

dewet22-codex:

custom_components/givenergy_local/time.py
```diff
@@ -101,6 +108,7 @@ class GivEnergyTimeEntityDescription(TimeEntityDescription):
         slot_fn=lambda inv: inv.battery_pause_slot_1,
         is_start=True,
         setter_fn=lambda value, _inv: commands.set_pause_slot_start(value),
+        skip_if_none=True,  # battery pause slot is firmware-gated (#207)
```
dewet22-codex: Red (blocking): `None` does not establish that HR319/320 are unreadable. The modbus `Converter.timeslot` deliberately returns `None` when either endpoint is the raw value `60`, which is the hardware sentinel for a valid but unset slot (the portal shows `--:--`); it also collapses malformed values to `None`. Gating the pause time descriptions on the decoded slot therefore removes both controls when the slot is merely unset, leaving the user no way to configure it. I think readability needs to be derived from register presence/freshness rather than the decoded `TimeSlot | None`, or the time controls need to remain present while only their state is unknown.

custom_components/givenergy_local/number.py
```diff
@@ -196,6 +205,30 @@ def _ems_number_descriptions() -> tuple[GivEnergyEmsNumberEntityDescription, ...
...
+        value = description.value_fn(inverter)
+    except Exception:  # noqa: BLE001
+        _LOGGER.warning("Skipping number %s: value_fn raised at setup", description.key)
+        return False
+    if value is None:
+        # Expected: the register isn't present on this device/firmware. DEBUG, not
+        # WARNING — this is the gate working as designed and recurs every restart,
+        # unlike the duplicate/garbled anomalies that warrant a warning.
+        _LOGGER.debug("Skipping number %s: register reads None at setup", description.key)
+        return False
```
dewet22-codex, line 234: Red (blocking): this only suppresses creation on fresh setup. On an upgraded entry, Home Assistant retains any existing HR313/314, pause-mode and pause-slot entity-registry rows when the platforms stop adding them, so users with unreadable registers keep the same orphaned/unavailable controls this PR is intended to remove. The EMS cleanup added for #205 handles the analogous transition explicitly. I think this needs a narrowly-scoped pre-platform registry reconciliation for the readability-gated unique IDs, with an upgrade regression that seeds old number/select/time rows and confirms they are removed while readable controls remain.

custom_components/givenergy_local/time.py
dewet22: Fixed in faa39ea — you're right that the decoded slot conflates "unset" (the raw-60 sentinel → None, shown as `--:--`) with "absent". Added a `readable_fn` presence signal to the time description and gated the pause slots on the pause-**mode** register (`inv.battery_pause_mode`, which is None only when the register is genuinely absent — no sentinel collapse), rather than `slot_fn`. Added a regression test: an unset slot (decodes to None) with the mode register present keeps both controls. Number/select are unaffected — their `None` genuinely means absent.

custom_components/givenergy_local/number.py
dewet22: Fixed in 6751f7d — you're right, suppressing creation leaves the orphaned rows on an upgraded entry. Added `_reconcile_readability_gated_controls` in `__init__.py` (pre-platform, mirroring the #205 EMS reconciliation): it removes rows for skip_if_none controls whose register is absent, reusing each platform's own gate (`_include_number`/`_include_select`/`_include_time`) so the readability logic isn't duplicated. Upgrade regression test seeds old number/select/time rows and asserts the unreadable ones are removed while a readable control (battery_power_mode) survives.

dewet22-codex:
I’m starting one final re-review of the latest changes now.

custom_components/givenergy_local/__init__.py
```diff
@@ -619,6 +661,10 @@ async def _on_topology_healed(confirmed: PlantCapabilities) -> None:
     if coordinator.data.ems is not None:
         _reconcile_ems_entities(hass, entry, coordinator.data.inverter_serial_number)
 
+    # Remove control rows whose register is absent on this device/firmware (#207):
```
dewet22-codex, line 671: Red (blocking): a cold-start partial poll is also represented by these accessors returning `None`, and the coordinator deliberately serves that partial with `last_partial_failures` set so failed-bank entities can recover on later polls. If HR(300) times out transiently on an AC/AIO setup, this path now removes existing HR313/314/pause rows and the platforms omit them; entity sets are frozen after setup, so a later clean poll cannot recover them without a reload. That turns a transient read failure into lost registry history/customisation plus missing controls. I think readability suppression/reconciliation needs to be disabled when the setup seed was partial (or, more precisely, when the relevant bank failed), retaining/creating the controls as unavailable until a clean read establishes structural absence. A cold-seed partial regression for the AC-config bank would pin this distinction.

dewet22: Fixed in the latest push — you're right, a partial seed conflates a transient bank failure with structural absence, and freezing entity sets after setup would make that loss unrecoverable until a reload. Threaded a `clean` flag (`not coordinator.last_partial_failures`) through the gate helpers so they don't suppress on a partial seed, and early-returned `_reconcile_readability_gated_controls` on a partial seed — structural absence is only concluded from a clean read. Added a clean-flag unit test and a partial-seed reconciliation no-op test.

dewet22-codex:
I’m starting the final re-review of the latest changes now.

dewet22-codex (approved):
I’ve completed the final re-review. The partial-seed fix now preserves both registry rows and platform entities whenever setup is degraded, while a clean absent read still performs the intended suppression and upgrade cleanup. The unset pause-slot case remains configurable through the separate pause-mode presence signal. The focused lifecycle suite (142 tests) and full suite (486 tests) pass locally. I have no blocking concerns remaining.

PR #210, opened 2026-06-24
fix: namespace EMS managed-inverter identity to prevent serial collision (#203)

dewet22:
## Summary
Fixes #203 — the EMS managed-inverter rollup devices (#199) collided by serial with a directly-connected inverter entry of the same serial.

Every device is identified as `(DOMAIN, serial)` with `unique_id = f"{serial}_{key}"`. The #199 rollup reused that shape with the managed inverter's serial, so running **both** an EMS controller entry and a direct connection to one of its managed inverters hit:
- **Device-identity hijack** — the rollup's `DeviceInfo` merged onto the real inverter's device, relabelling it "Managed Inverter (EMS)" and reparenting it under the controller.
- **Dropped sensors** — the rollup's `status`/`battery_soc` collided on `unique_id` and were dropped (2 of 4 rollup sensors vanished).

## Fix
Namespace the rollup device + entity identity with a `_managed` marker: `(DOMAIN, f"{serial}_managed")` / `f"{serial}_managed_{key}"`. Serials never contain `_managed`, so it can no longer collide; the real serial still shows via the device name and `serial_number`.

## Migration
No registry migration. The old `{serial}_{key}` ids are *ambiguous* — for a user with a direct entry they belong to the real inverter — so a blind rename would clobber the wrong entity (that ambiguity is the bug). Instead it self-heals: once the rollup stops claiming `(DOMAIN, serial)`, the direct inverter re-registers that device with its correct name/model on restart, leaving only the rollup's old rows orphaned (clearable via the device page). #199 shipped a day ago (v1.3.9) to very few EMS users, so the blast radius is minimal.

## Tests
- Updated the managed-inverter device/sensor tests to the `_managed` shape.
- Added `test_managed_inverter_namespaced_against_same_serial` — the in-one-entry repro (rollup reports the controller's own serial): the controller keeps its bare-serial identity and `status` sensor while the rollup gets a distinct `_managed` device + namespaced sensors.

487 passed; mypy + ruff clean.

## Release Notes

* **Bug Fixes**
  * Fixed entity identity collisions when managed inverter rollup sensors share the same serial number with directly-connected inverter sensors. Each device now maintains a distinct identity in Home Assistant.

* **Tests**
  * Added regression test to verify managed inverters do not hijack identities of same-serial direct devices.

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -2221,7 +2221,9 @@ class GivEnergyManagedInverterSensor(CoordinatorEntity[GivEnergyUpdateCoordinato
     manages appear only as blinded rollup summaries (status / power / SoC /
     heatsink temp — no per-string detail). Each managed inverter becomes its own
     HA device, parented to the controller via `via_device` and identified by its
-    own serial. The rollup lives in the EMS block on 0x11 (there's no per-inverter
+    own serial namespaced with a `_managed` marker — so the rollup can't collide
+    with a directly-connected inverter entry of the same serial (#203). The rollup
+    lives in the EMS block on 0x11 (there's no per-inverter
     register bank), so there's no stale gate: availability is simply whether the
     serial is still present in the latest poll.
     """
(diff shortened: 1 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve completed the review and have no blocking concerns. The managed-device and entity namespace prevents both sides of the serial collision, while retaining the real serial as device metadata. I also checked the upgrade case: when the directly connected entity is recreated, Home Assistant transfers an existing ambiguous registry row to that config entry and device, so the documented restart-based recovery is consistent with the registry behaviour. The EMS-focused tests (30) and full suite (487) pass locally, and the required CI checks are green.

PR #211, opened 2026-06-24
fix: target the EMS controller on the auto-dashboard (#175)

### LINKED ISSUE #175: Auto generated dashboard not registering both inverters in EMS setup

dewet22:
## Summary
Fixes #175 — on an EMS plant the auto-generated dashboard recognised only one inverter and showed zero PV.

The dashboard is a Lovelace **strategy in JS** (`ge-strategy.js`). `buildPlant` pooled the EMS controller with every inverter and defaulted `target = inverters[0]` (alphabetical), so an EMS plant landed on an arbitrary managed inverter reporting no PV of its own. Even when it hit the controller, `classicViews` short-circuited EMS to `["EMS Controls", "Diagnostics"]` with no telemetry. #206 has since restored the controller's own PV/grid/battery/AC sensors, so the data is present on the controller.

## Fix (Level 1)
- **Target the EMS controller** when no `serial` is pinned (the pin still overrides).
- **Telemetry views for the controller** — Overview + Energy — then append EMS Controls + Diagnostics (batteries hang off the managed inverters; inverter-level controls are suppressed on EMS).
- **EMS-aware load accessor** `a.load()` = `ems_calc_load_power || p_load_demand` — `p_load_demand` is gated off on the controller (#206), so the power-flow home slot + 24h Load series use the EMS aggregate; plain plants are unchanged.
- **"EMS Plant" card** on the Overview surfacing the controller's aggregates (calc load, grid meter, total battery, remaining energy, managed-inverter count; measured load omitted — reads zero on current firmware).

Per-mode views (flow/glance/analyst/all) keep falling back to the now-useful classic EMS set — a deliberate Level-1 boundary.

## Out of scope (follow-ups)
- **Level 2:** a "Managed Inverters" view enumerating the `_managed` devices.
- Nick's other #52 items (control bugs + Predbat sensors) — separate issues.

## Tests
- `mock-hass.js`: the `ems:true` controller now models its post-#206 keys (inverter sensors minus the gated/suppressed ones, plus the EMS aggregates).
- `ge-strategy.test.js`: rewrote the 5 EMS cases to the new view set + Overview content, and added target-selection tests (controller preferred over a lower-sorting plain inverter; serial pin still honoured).
- 40 JS tests, parity guard, 487 Python tests, ruff + mypy — all green. Rendering needs Nick's plant to confirm.

### DIFF
--- a/custom_components/givenergy_local/__init__.py
+++ b/custom_components/givenergy_local/__init__.py
@@ -71,7 +71,7 @@ _STRATEGY_URL = f"/{DOMAIN}/{_STRATEGY_FILENAME}"
 # served from the same package dir so they resolve offline without a CDN.
 _FONTS_DIRNAME = "fonts"
 _FONTS_URL = f"/{DOMAIN}/{_FONTS_DIRNAME}"
-_STRATEGY_VERSION = "9"
+_STRATEGY_VERSION = "10"
 
 # Per-config-entry topology cache. PlantCapabilities is persisted as
 # `to_dict()` directly (no envelope) following HA Core's Store convention —
--- a/custom_components/givenergy_local/www/ge-strategy.js
+++ b/custom_components/givenergy_local/www/ge-strategy.js
@@ -173,7 +173,13 @@
         }) || null;
       if (!target) unmatchedSerial = opts.serial;
     } else {
-      target = inverters[0] || null;
+      // On an EMS plant prefer the controller (it carries the plant-level
+      // aggregates and the real PV/grid/battery telemetry); a plain multi-inverter
+      // plant still takes the first inverter by serial.
+      var emsTarget = inverters.find(function (p) {
+        return p.isEms;
+      });
+      target = emsTarget || inverters[0] || null;
     }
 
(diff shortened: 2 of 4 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve completed the review and have no blocking concerns. The registry classification keeps managed rollup devices out of the inverter pool, prefers the EMS controller only when no serial is pinned, and preserves explicit targeting of an independent plain inverter. The controller’s post-#206 telemetry and EMS aggregate keys line up with the generated Overview and Energy configuration, with suppressed fields omitted rather than left dangling. The frontend suite (40 tests), strategy/EMS parity slice (32 tests), and full Python suite (487 tests) pass locally; all CI checks are green. I have not been able to confirm the final visual composition against the live EMS plant, as already noted in the PR.

PR #213, opened 2026-06-24
fix: EMS managed-inverter status crash + measured-load hide + AC-limit rename (#52)

dewet22:
## Summary
Three hass-local fixes from Nick's #52 EMS cross-check, as atomic commits.

### 1. Managed-inverter Status sensor crash (the priority)
On an EMS plant the controller reports each managed inverter's status as a raw string, not a `Status` enum, so the rollup status sensor's `summary.status.name` raised `AttributeError: 'str' object has no attribute 'name'` on every poll — 350 occurrences in Nick's logs. A small `_managed_status` helper now coerces defensively (`.name` only when present, else `str(value)`). The existing tests passed only because the fixture defaulted `status=None`; added a regression with a string status.

### 2. Hide always-zero EMS Measured Load Power
The controller reports a constant zero for measured load on current firmware, where the calculated-load aggregate carries the real figure. Hidden by default on new installs (still recorded, matching the existing `grid_power` precedent) to declutter the default view. The flag only applies at first entity registration — existing installs keep the entity visible until hidden manually.

### 3. Rename the AC charge/discharge limits to the app names
HR313/314 are "Inverter Charge Power Percentage" / "Inverter Discharge Power Percentage" in the GivEnergy app; the integration called them "Battery AC Charge/Discharge Limit". The keys (and thus unique_ids / existing entity_ids) are unchanged, so existing installs keep their entity_ids and just see the new friendly name.

## Out of scope (routed/deferred)
- **DC battery limit 0–50 → 0–100 (HR111/112):** blocked on the modbus library, whose `set_battery_charge_limit`/`set_battery_discharge_limit` hard-raise outside 0–50. Filing a givenergy-modbus issue to widen the range; the one-line hass `native_max_value` bump follows once that releases and is pinned.
- The remaining #52 items (Export Priority read mismatch, Max Output Active Power write no-op, Predbat helper sensors, config-flow battery-only reconfigure, managed-inverter controls) — separate passes.

## Tests
- New regression for the string-status crash; guard test that Measured Load Power ships hidden.
- 489 Python + 40 JS tests, parity guard, mypy, ruff — all green. The rename touches only `name=`, so no entity-id churn and no parity-guard impact.

Rendering of the managed-inverter status on real EMS hardware still wants Nick's confirmation once v1.3.17 is out.

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -1391,9 +1391,10 @@ EMS_SENSORS: tuple[GivEnergyEmsSensorDescription, ...] = (
         device_class=SensorDeviceClass.POWER,
         state_class=SensorStateClass.MEASUREMENT,
         value_fn=_ems_attr("measured_load_power"),
-        # Reads a constant zero on current EMS firmware; the calculated load
-        # aggregate supersedes it. Hidden by default but still recorded, so it
-        # surfaces automatically if a future firmware starts populating it (#52).
+        # Reads a constant zero on current EMS firmware, where the calculated-load
+        # aggregate carries the real figure. Hidden by default on new installs
+        # (still recorded, like grid_power above) to declutter; existing installs
+        # keep it until hidden manually — the flag only applies at registration (#52).
         entity_registry_visible_default=False,
     ),
     GivEnergyEmsSensorDescription(
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -1391,6 +1391,10 @@ EMS_SENSORS: tuple[GivEnergyEmsSensorDescription, ...] = (
         device_class=SensorDeviceClass.POWER,
         state_class=SensorStateClass.MEASUREMENT,
         value_fn=_ems_attr("measured_load_power"),
+        # Reads a constant zero on current EMS firmware; the calculated load
+        # aggregate supersedes it. Hidden by default but still recorded, so it
+        # surfaces automatically if a future firmware starts populating it (#52).
+        entity_registry_visible_default=False,
     ),
     GivEnergyEmsSensorDescription(
         key="ems_grid_meter_power",
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -148,6 +148,19 @@ def _summary_attr(name: str) -> Callable[[InverterSummary], Any]:
     return lambda summary: getattr(summary, name)
 
 
+def _managed_status(value: Any) -> str | None:
+    """Render an EMS managed inverter's status defensively.
+
+    Unlike a directly-reachable inverter's Status enum, the EMS rollup reports
+    each managed inverter's status as a raw string, so coerce instead of
+    assuming `.name` — calling `.name` on the string raised on every poll on
+    real EMS hardware (#52).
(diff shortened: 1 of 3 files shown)


dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -1378,6 +1391,10 @@ def _ems_attr(name: str) -> Callable[[Ems], Any]:
         device_class=SensorDeviceClass.POWER,
         state_class=SensorStateClass.MEASUREMENT,
         value_fn=_ems_attr("measured_load_power"),
+        # Reads a constant zero on current EMS firmware; the calculated load
+        # aggregate supersedes it. Hidden by default but still recorded, so it
+        # surfaces automatically if a future firmware starts populating it (#52).
+        entity_registry_visible_default=False,
```
dewet22-codex, line 1398: Red (blocking): this flag is only consulted when Home Assistant creates the registry row. `EntityRegistry.async_get_or_create()` does not update `hidden_by` for an existing entity, so users upgrading from the release that introduced this sensor will continue to see the zero-valued row. Conversely, a fresh install gets `hidden_by=integration` permanently; a later non-zero value does not make it resurface automatically. The descriptor test only proves the static default, not either lifecycle claim. I think this needs either registry reconciliation with clearly defined user-override handling, or a narrower fresh-install-only scope with the automatic-resurfacing and upgrade claims removed.

dewet22-codex (changes requested):
I found one blocking lifecycle mismatch in the measured-load visibility change. The registry visibility default applies only when an entity is first created: upgraded installs keep the existing row visible, while fresh installs keep it integration-hidden even if future firmware starts reporting a useful value. I’ve left the Home Assistant registry behaviour and possible scopes inline. The managed-status crash fix and the AC-limit friendly-name changes look sound. The full Python suite passes locally (489 tests); all 40 frontend tests also pass, with the local command only failing afterward when Vitest attempted to write its cache into the read-only checkout. CI is otherwise green.

custom_components/givenergy_local/sensor.py
dewet22: Fair point — the comment has been narrowed to accurately describe fresh-install-only scope. The new text reads:

> Reads a constant zero on current EMS firmware, where the calculated-load aggregate carries the real figure. Hidden by default on new installs (still recorded, like grid_power above) to declutter; existing installs keep it until hidden manually — the flag only applies at registration (#52).

The automatic-resurfacing claim is gone. Existing installs keep the entity visible until the user hides it manually, which is standard HA behaviour and acceptable here.

dewet22-codex:
I’m starting a re-review of the measured-load lifecycle fix now.

dewet22-codex (approved):
I’ve re-reviewed the amended head and the blocking lifecycle concern is resolved. The implementation is now explicitly scoped to fresh-install default visibility: existing registry rows remain user-controlled, and no automatic resurfacing is claimed in the code. The managed-status handling and AC-limit renames remain sound. The full Python suite (489 tests), focused EMS/number/parity suite (46 tests), and frontend suite (40 tests) pass locally; all CI checks are green. One non-blocking documentation note remains: the PR description still says the measured-load sensor “resurfaces automatically if a firmware starts populating it”, which no longer matches the deliberately narrowed scope.

PR #214, opened 2026-06-24
fix(deps): bump givenergy-modbus to 2.5.8, widen DC + AC battery limits to 0–100

dewet22:
## Summary
Bumps the modbus pin to 2.5.8 and gets the battery power-limit controls right for every plant topology. Three commits:

**1. Bump to 2.5.8, widen both limit pairs to 0–100.** 2.5.7 fixed a splice-guard false positive that froze battery SOC/cell sensors near full charge; 2.5.8 widens the DC (HR111/112) and AC (HR313/314) charge/discharge setters to accept 0–100 (modbus #301/#302). The DC sliders (capped at 50) and AC sliders (1-floor) both move to 0–100 in lockstep with the setter contract.

**2. Route AC-coupled battery control to the AC pair.** On plants exposing the AC-config block (AC-coupled inverters and single-phase All-in-One), battery power is controlled via the AC pair (HR313/314); the DC pair (HR111/112) targets a different register there and was misleadingly shown alongside it. The DC battery-limit controls are now suppressed on those plants. DC-coupled hybrids are unaffected. *Effect on existing AC-coupled/AIO installs: the `Battery Charge/Discharge Limit` entities are removed (wrong register); the `Inverter Charge/Discharge Power Percentage` controls remain.*

**3. Boundary write tests** for charge and discharge, DC and AC, at 0 and 100 — guarding that the slider range never advertises a write the library rejects.

## Follow-up (hardware verification)
modbus dropped the AC pair's 1-floor to 0, but the "0 disables charge/discharge" semantics on HR313/314 are unverified (the firmware doc is hybrid-scoped). Asking Nick to write 0 to HR313/314 on his AC/EMS plant and confirm it disables AC charge/discharge — closes the loop on the AC 0-floor (modbus #302).

## Tests
500 Python + 40 JS tests, parity guard, mypy, ruff — all green.

### DIFF
--- a/custom_components/givenergy_local/number.py
+++ b/custom_components/givenergy_local/number.py
@@ -109,6 +109,11 @@ NUMBER_DESCRIPTIONS: tuple[GivEnergyNumberEntityDescription, ...] = (
 
 # --- AC-config-block controls (AC-coupled inverters + single-phase All-in-One) ---
 
+# DC battery power-limit controls (HR111/112) suppressed on plants that expose the
+# AC-config block — there the AC pair (HR313/314) is the battery-power control and the
+# DC pair targets a different register (modbus #301/#302).
+_DC_BATTERY_LIMIT_KEYS = frozenset({"battery_charge_limit", "battery_discharge_limit"})
+
 # The AC charge/discharge power limits (HR313/314) are distinct from the DC-side
 # limits above (HR111/112) and are only meaningful on models that expose the
 # AC-config register block (HR300+). Gated via PlantCapabilities.has_ac_config_block
@@ -253,11 +258,22 @@ async def async_setup_entry(
             for description in EMS_NUMBER_DESCRIPTIONS
         )
     else:
-        entities.extend(
-            GivEnergyNumberEntity(coordinator, description) for description in NUMBER_DESCRIPTIONS
-        )
         caps = coordinator.data.capabilities
-        if caps is not None and caps.has_ac_config_block and not caps.is_three_phase:
+        ac_battery_control = (
+            caps is not None and caps.has_ac_config_block and not caps.is_three_phase
+        )
+        descriptions: tuple[GivEnergyNumberEntityDescription, ...] = NUMBER_DESCRIPTIONS
+        if ac_battery_control:
+            # On AC-coupled / AIO plants battery power is controlled via the AC pair
+            # (HR313/314) created below; the DC pair (HR111/112) targets a different
+            # register here and would mislead, so suppress it (modbus #301/#302).
+            descriptions = tuple(
+                d for d in NUMBER_DESCRIPTIONS if d.key not in _DC_BATTERY_LIMIT_KEYS
+            )
+        entities.extend(
+            GivEnergyNumberEntity(coordinator, description) for description in descriptions
+        )
+        if ac_battery_control:
             inverter = coordinator.data.inverter
             clean = not coordinator.last_partial_failures
             entities.extend(
--- a/custom_components/givenergy_local/number.py
+++ b/custom_components/givenergy_local/number.py
@@ -61,7 +61,7 @@ NUMBER_DESCRIPTIONS: tuple[GivEnergyNumberEntityDescription, ...] = (
         name="Battery Charge Limit",
         native_unit_of_measurement=PERCENTAGE,
         native_min_value=0,
-        native_max_value=50,
+        native_max_value=100,
         native_step=1,
         mode=NumberMode.BOX,
         value_fn=lambda inv: inv.battery_charge_limit,
@@ -73,7 +73,7 @@ NUMBER_DESCRIPTIONS: tuple[GivEnergyNumberEntityDescription, ...] = (
         name="Battery Discharge Limit",
         native_unit_of_measurement=PERCENTAGE,
         native_min_value=0,
-        native_max_value=50,
+        native_max_value=100,
         native_step=1,
         mode=NumberMode.BOX,
         value_fn=lambda inv: inv.battery_discharge_limit,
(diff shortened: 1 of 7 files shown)


dewet22-codex:

custom_components/givenergy_local/number.py
```diff
@@ -61,7 +61,7 @@ class GivEnergyNumberEntityDescription(NumberEntityDescription):
         name="Battery Charge Limit",
         native_unit_of_measurement=PERCENTAGE,
         native_min_value=0,
-        native_max_value=50,
+        native_max_value=100,
```
dewet22-codex, line 64: Red (blocking): the pinned `givenergy-modbus==2.5.7` still raises `ValueError` from both command builders above 50, and `async_set_native_value()` evaluates `set_value_cmd` without catching that exception. This therefore advertises 51–100 as valid Home Assistant values while every write in that range fails before a request is sent. I don’t think the UI range can move ahead of the write contract: either keep these maxima at 50 in this PR, or pin the first modbus release whose charge/discharge setters accept 0–100 and add a boundary write test here.

dewet22-codex (changes requested):
I found one blocking cross-repository contract mismatch. The new controls expose 0–100, but the pinned 2.5.7 command builders still reject 51–100 before any device write, and the entity write path does not absorb that validation error. I’ve left the concrete call-path detail inline. The 2.5.7 dependency bump and lockfile update otherwise look clean; the focused number tests pass.

custom_components/givenergy_local/number.py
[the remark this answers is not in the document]
dewet22: Resolved by pinning `givenergy-modbus>=2.5.8`, whose DC (HR111/112) and AC (HR313/314) charge/discharge setters now accept the full 0–100 range (modbus #301/#302). The UI range and the write contract move together — added boundary write tests covering 0 and 100 against both pairs.

custom_components/givenergy_local/number.py
```diff
@@ -73,7 +73,7 @@ class GivEnergyNumberEntityDescription(NumberEntityDescription):
         name="Battery Discharge Limit",
         native_unit_of_measurement=PERCENTAGE,
         native_min_value=0,
-        native_max_value=50,
+        native_max_value=100,
```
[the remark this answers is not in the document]
dewet22: Resolved by pinning `givenergy-modbus>=2.5.8`, whose DC (HR111/112) and AC (HR313/314) charge/discharge setters now accept the full 0–100 range (modbus #301/#302). The UI range and the write contract move together — added boundary write tests covering 0 and 100 against both pairs.

custom_components/givenergy_local/number.py
dewet22: Resolved by pinning `givenergy-modbus>=2.5.8`, whose DC (HR111/112) and AC (HR313/314) charge/discharge setters now accept the full 0–100 range (modbus #301/#302). The UI range and the write contract move together — added boundary write tests covering 0 and 100 against both pairs.

custom_components/givenergy_local/number.py
```diff
@@ -113,14 +113,14 @@ class GivEnergyNumberEntityDescription(NumberEntityDescription):
 # limits above (HR111/112) and are only meaningful on models that expose the
 # AC-config register block (HR300+). Gated via PlantCapabilities.has_ac_config_block
 # (and not is_three_phase — three-phase AC remaps the read-back to different registers
-# than the command writes; see modbus#75). The setter rejects values below 1, hence the
-# 1–100 range.
+# than the command writes; see modbus#75). Range is 0–100% to match the GivEnergy
+# app and the DC-side pair.
```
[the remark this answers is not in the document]
dewet22: Leaving as-is. This project's ruff config selects `E/F/I/UP`, not `RUF`, so RUF003 isn't part of the lint gate (`ruff check` passes clean), and en-dashes are the established style throughout these comments — changing only this line would leave it inconsistent with the rest of the file.

dewet22-codex:
I’m starting a fresh review of the current head now.

custom_components/givenergy_local/number.py
```diff
@@ -253,11 +258,22 @@ async def async_setup_entry(
             for description in EMS_NUMBER_DESCRIPTIONS
         )
     else:
+        caps = coordinator.data.capabilities
+        ac_battery_control = (
+            caps is not None and caps.has_ac_config_block and not caps.is_three_phase
+        )
+        descriptions: tuple[GivEnergyNumberEntityDescription, ...] = NUMBER_DESCRIPTIONS
+        if ac_battery_control:
```
dewet22-codex, line 266: Red (blocking): this suppresses the DC pair only for fresh entity enumeration. On an upgraded AC/AIO entry, Home Assistant keeps the existing `battery_charge_limit` and `battery_discharge_limit` registry rows when the platform stops adding them — the integration already handles this exact invariant in `_reconcile_ems_entities()` and `_reconcile_readability_gated_controls()`. Those stale rows remain enabled, and the bundled strategy resolves controls from the entity registry, so existing dashboards can still render the wrong DC controls as unavailable alongside the AC pair. I think this topology change needs the matching pre-platform reconciliation for these two unique IDs, with an upgrade test that seeds the old rows and confirms AC/AIO removes them while a hybrid retains them.

dewet22-codex (changes requested):
Taking the current head afresh, I found one blocking upgrade-path gap in the new topology routing. Fresh AC/AIO installs omit the wrong DC controls correctly, but upgraded entries retain their existing registry rows unless they are explicitly reconciled; the bundled dashboard can therefore continue to surface those stale controls. I’ve left the concrete lifecycle detail and test scope inline. The original cross-repository range mismatch is resolved by the 2.5.8 pin, all four setter boundaries are covered, `uv lock --check` is clean, focused lint passes, and the full Python suite passes (500 tests).

custom_components/givenergy_local/number.py
dewet22: Good catch — the suppression only covered fresh enumeration. Added `_reconcile_ac_coupled_dc_limits`, mirroring `_reconcile_ems_entities` / `_reconcile_readability_gated_controls`: on plants with the AC-config block it removes the stale `battery_charge_limit` / `battery_discharge_limit` rows pre-platform (between the readability reconcile and `async_forward_entry_setups`). It's gated on plant capability rather than a register read, so — like the EMS reconciler — it needs no partial-poll guard. Upgrade tests seed the old DC rows and confirm AC/AIO removes them while a DC-coupled hybrid retains them.

dewet22-codex:
I’m starting a final review of the latest changes now.

dewet22-codex (approved):
I’ve completed the final review and have no blocking concerns remaining. The upgrade-path gap is resolved: stale DC limit rows are removed before platform setup only for plants that positively expose the single-phase AC-config block, while DC-coupled hybrids retain them. The 2.5.8 setter contract, all four 0/100 boundary paths, fresh-install topology and upgrade reconciliation are covered. The focused init/number suite passes (67 tests), the full suite passes (502 tests), lint and lock checks are clean, and CI is green.

PR #215, opened 2026-06-24
chore(ci): drop the retired givenergy-modbus-release dispatch trigger

dewet22:
## Summary
givenergy-modbus retired the downstream-notify `repository_dispatch` (type `givenergy-modbus-release`) that fired this bump workflow — modbus now announces releases via the cross-agent coordination channel instead. The listener here was dead.

This removes the `repository_dispatch` trigger and the dispatch-specific plumbing, keeping the workflow on `workflow_dispatch` only:
- Drop the `repository_dispatch: [givenergy-modbus-release]` trigger; update the header comment.
- Remove the `client_payload.version` env and the `${DISPATCH_VERSION:-$MANUAL_VERSION}` fallback — `NEW` is now just `inputs.version`.
- Reframe the trust-boundary comment around the `workflow_dispatch` input; the PEP 440 input validation is unchanged and now guards that single boundary.

The manual path is retained and unchanged in behaviour — it's still the way to bump the pin by hand (e.g. around the PyPI propagation race: `gh workflow run bump-givenergy-modbus.yml -f version=X.Y.Z`). `BUMP_PAT` stays, since manual runs still push the branch and open the PR so `validate.yml` fires.

Net effect: removing the dead trigger also drops the only `client_payload.*` (attacker-controllable repository_dispatch field) usage in the file.

## Summary by CodeRabbit

* **Chores**
  * Updated the release workflow to use manual version entry only.
  * Tightened version handling to better protect against invalid or unsafe input.
  * Removed the previous cross-repo trigger and refreshed the workflow notes accordingly.

### DIFF
--- a/.github/workflows/bump-givenergy-modbus.yml
+++ b/.github/workflows/bump-givenergy-modbus.yml
@@ -1,10 +1,10 @@
 name: Bump givenergy-modbus
 
-# Triggered by a repository_dispatch event from dewet22/givenergy-modbus's
-# release workflow. Also exposed as workflow_dispatch for manual runs.
+# Run manually to bump the givenergy-modbus pin and open a PR. modbus releases
+# are announced via the cross-agent coordination channel (the automated
+# repository_dispatch from givenergy-modbus's release workflow was retired), so
+# this is workflow_dispatch only.
 on:
-  repository_dispatch:
-    types: [givenergy-modbus-release]
   workflow_dispatch:
     inputs:
       version:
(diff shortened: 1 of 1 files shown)


dewet22-codex:
I’m starting a quick review of the current changes now.

dewet22-codex (approved):
Quick review complete: the retired dispatch trigger and its payload plumbing are removed cleanly, while the manual workflow retains the same required input, validation, branch routing, PAT-authenticated push and PR creation behaviour. I found no blocking concerns. The head is mergeable and all CI checks are green.

PR #216, opened 2026-06-24
fix: invert the power-flow-card-plus grid sign (#212)

dewet22:
## Summary
Fixes the Overview power-flow card showing grid **import as export** (Nick, #212).

The `custom:power-flow-card-plus` grid bubble was fed the raw signed `grid_power` sensor, which is **+ve = export** (GE convention). `power-flow-card-plus` treats a single grid entity as **+ve = import/consumption**, so the sign was backwards — an importing plant rendered as exporting, and the home bubble's flow was thrown off as a result.

The card exposes `invert_state`, which flips both the direction and the value. Setting it on the grid entity makes the card's convention match the sensor.

```js
ents.grid = { entity: a.inv("grid_power"), invert_state: true };
```

## Not EMS-specific
The same unguarded `grid_power` feeds the flow card on every plant type — Nick just hit it on the EMS Overview. The bundled custom flow (`givenergy-flow`) already decodes the sign correctly (+ = export, − = import); only this third-party card was unguarded. Battery and solar are left as-is — they render correctly and weren't reported.

## Changes
- `ge-strategy.js` — `invert_state: true` on the grid entity.
- `__init__.py` — `_STRATEGY_VERSION` 10 → 11 (cache-bust so browsers fetch the corrected strategy).
- `ge-strategy.test.js` — regression test asserting the grid entity carries `invert_state`.

## Tests
502 Python + 41 JS (incl. the new regression), all green.

## Out of scope
The per-battery SOC / charge-status feature request in the same issue is tracked separately.

### DIFF
--- a/custom_components/givenergy_local/__init__.py
+++ b/custom_components/givenergy_local/__init__.py
@@ -71,7 +71,7 @@ _STRATEGY_URL = f"/{DOMAIN}/{_STRATEGY_FILENAME}"
 # served from the same package dir so they resolve offline without a CDN.
 _FONTS_DIRNAME = "fonts"
 _FONTS_URL = f"/{DOMAIN}/{_FONTS_DIRNAME}"
-_STRATEGY_VERSION = "10"
+_STRATEGY_VERSION = "11"
 
 # Per-config-entry topology cache. PlantCapabilities is persisted as
 # `to_dict()` directly (no envelope) following HA Core's Store convention —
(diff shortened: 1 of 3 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the current head and have no blocking concerns. The integration’s `grid_power` contract is positive for export, while power-flow-card-plus documents a single grid entity as positive for consumption; its `invert_state` option explicitly reverses both direction and displayed value, so this is the correct adapter. The strategy-version bump ensures existing browsers fetch the change, and the regression pins the generated card config. All 41 frontend tests exercised successfully locally (Vitest only failed afterward trying to write its cache in the read-only checkout), the focused Python tests pass (44), and CI is green.

PR #222, opened 2026-06-25
feat: experimental-features mechanism for opt-in modbus client flags

dewet22:
## Summary
Adds a registry-driven **"Experimental features"** options group so opt-in givenergy-modbus *client* flags can be toggled from Home Assistant and forwarded into the `Client(...)` construction. Built ahead of a 2.5.11 opt-in feature so the first flag is a one-line addition.

The mechanism mirrors the modbus client's own `splice_heal_seconds` idiom (an off-by-default optional kwarg acted on only when set), and reuses the existing `CONF_BATTERY_DATA_ONLY` config/options plumbing plus the options-reload listener.

## How it works
- **`const.py`** — `ExperimentalFeature` frozen dataclass + `EXPERIMENTAL_FEATURES` registry (**ships empty**) + `resolve_experimental_client_kwargs(options)`. The resolver reads the nested options section and returns `{}` when nothing is enabled, so an all-off entry constructs `Client(host, port)` byte-identically to before.
- **`config_flow.py`** — a collapsed HA `section()` in the OptionsFlow, added **only when the registry is non-empty**, so no empty "Experimental features" header appears before any flag ships. Section values persist nested under `entry.options["experimental"]`.
- **`__init__.py` / `coordinator.py`** — resolve from `entry.options` at setup and splat the kwargs into the single live `Client(...)` call. The existing `add_update_listener` reload already rebuilds the client on any options change — no new reload wiring.

## Zero user-facing effect today
The registry is empty, so the section never renders and `Client(...)` is unchanged. **Adding the first flag** = one `ExperimentalFeature(...)` registry line + its `strings.json`/`translations` row + (when the kwarg ships) bumping the `givenergy-modbus` pin — all in a single commit. No runtime version-guard is needed because the pin bump is atomic with the kwarg-passing entry. Translations are intentionally deferred to that commit (the section doesn't render while empty, so adding section copy now would be unused/hassfest-flaggable).

## Decisions
- **Options-only** (post-setup Configure), not initial setup — experimental client flags only affect transport and take effect on reload, so there's no first-run-only window (unlike `battery_data_only`, which gates entity creation).
- **`client_value: Any`** (not bool-only) on the registry — the template kwarg `splice_heal_seconds` is a float; the UI toggle stays boolean, `client_value` is what's sent when on.

## Tests
520 Python tests green (+12 new): resolver invariants (default-off, unknown-key-ignored, enabled→kwarg, bool default-True), options section (renders/persists/prefills with a patched one-feature registry; **omitted** when empty), and client-kwarg forwarding (enabled flag splatted; default-off passes only host/port). ruff, mypy, prek all clean.

## Summary by CodeRabbit

* **New Features**
  * Added an optional **Experimental features** section in the options flow, with settings that can be enabled or preserved across saves.
  * Experimental settings now carry through to the integration setup automatically.

* **Bug Fixes**
  * Improved handling of saved options so previously enabled experimental settings are not lost when re-opening the configuration.

* **Chores**
  * Updated the bundled `givenergy-modbus` dependency to a newer compatible version.

### DIFF
--- a/custom_components/givenergy_local/config_flow.py
+++ b/custom_components/givenergy_local/config_flow.py
@@ -173,10 +173,18 @@ class GivEnergyLocalOptionsFlow(OptionsFlow):
         # Surface the collapsed "Experimental features" group only once at least one
         # flag exists, so the header never appears empty (the registry ships empty).
         if EXPERIMENTAL_FEATURES:
-            schema_dict[vol.Optional(CONF_EXPERIMENTAL, default={})] = section(
+            # Use currently-saved values as the schema default for each feature so
+            # that a collapsed/omitted section (which HA fills with the default
+            # rather than re-submitting the current values) falls back to what was
+            # already enabled rather than all-off feature defaults.
+            existing_exp: dict[str, Any] = self.config_entry.options.get(CONF_EXPERIMENTAL, {})
+            schema_dict[vol.Optional(CONF_EXPERIMENTAL, default=existing_exp)] = section(
                 vol.Schema(
                     {
-                        vol.Required(feature.conf_key, default=feature.default): bool
+                        vol.Required(
+                            feature.conf_key,
+                            default=existing_exp.get(feature.conf_key, feature.default),
+                        ): bool
                         for feature in EXPERIMENTAL_FEATURES
                     }
                 ),
--- a/custom_components/givenergy_local/config_flow.py
+++ b/custom_components/givenergy_local/config_flow.py
@@ -14,9 +14,11 @@ from homeassistant.config_entries import (
 )
 from homeassistant.const import CONF_HOST, CONF_PORT
 from homeassistant.core import callback
+from homeassistant.data_entry_flow import SectionConfig, section
 
 from .const import (
     CONF_BATTERY_DATA_ONLY,
+    CONF_EXPERIMENTAL,
     CONF_PASSIVE,
     CONF_SCAN_INTERVAL,
     DEFAULT_BATTERY_DATA_ONLY,
@@ -24,6 +26,7 @@ from .const import (
     DEFAULT_PORT,
     DEFAULT_SCAN_INTERVAL,
     DOMAIN,
+    EXPERIMENTAL_FEATURES,
 )
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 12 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the current head and have no blocking concerns. The empty registry leaves today’s client construction unchanged; nested section values round-trip through Home Assistant’s suggested-value handling; only registry-approved keys can become client kwargs; and the existing options reload plus coordinator reconnect path rebuilds the client consistently. The full suite (520 passed), Ruff, mypy, hassfest, and CI are clean.

tests/test_config_flow.py
```diff
@@ -254,3 +254,57 @@ async def test_options_flow_prefills_existing_value(hass, mock_client, setup_int
...
+
+
+# --- Experimental features section (client feature-flagging) ------------------
+
+from unittest.mock import patch  # noqa: E402
+
+from custom_components.givenergy_local.const import (  # noqa: E402
+    CONF_EXPERIMENTAL,
+    ExperimentalFeature,
+)
```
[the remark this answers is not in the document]
dewet22: Fixed — imports moved to the top of the file in 1df9bc9.

dewet22-codex:

custom_components/givenergy_local/config_flow.py
```diff
@@ -159,12 +162,27 @@ class GivEnergyLocalOptionsFlow(OptionsFlow):
 
     async def async_step_init(self, user_input: dict[str, Any] | None = None) -> ConfigFlowResult:
         if user_input is not None:
+            # user_input carries battery_data_only flat, plus — when any experimental
+            # feature is registered — the section's toggles nested under
+            # CONF_EXPERIMENTAL. Persisted as-is; resolve_experimental_client_kwargs
+            # reads that nested shape at coordinator construction.
             return self.async_create_entry(data=user_input)
```
dewet22-codex, line 169: [Red / blocking] Now that a real flag can be enabled, accepting an omitted experimental section by persisting `user_input` as-is can silently turn it off. Options-flow data replaces the entry’s whole options mapping: if the entry currently has `experimental: {splice_reject_heal: true}` and the collapsed section is omitted while the user changes only `battery_data_only`, this writes no experimental value (or the schema default `{}`) and the next reload disables the client kwarg. I think an omitted section needs to preserve the existing nested options, while an explicitly submitted false value should still disable it; a regression test for that sequence would pin the distinction.

dewet22-codex (changes requested):
The first real experimental flag is wired consistently to givenergy-modbus 2.5.11, but I found one blocking persistence issue: an omitted collapsed section is saved as absent/empty and replaces the previous nested options, silently disabling an enabled flag when another option is changed. I’ve left the sequence inline. The focused tests (132 passed), Ruff, lockfile check, and CI are otherwise clean.

custom_components/givenergy_local/config_flow.py
dewet22: Fixed in 3bc1ffd: instead of post-submit merging, the schema default for the section (and each inner feature) is seeded from the currently-saved options at render time. An omitted collapsed section now round-trips the live values rather than all-off defaults. Regression test added.

PR #225, opened 2026-06-25
fix: suppress Battery Charge/Discharge Today on EMS controllers (#221)

dewet22:
Fixes #221 (in part — the "unknown" half).

On an EMS plant, **Battery Charge Today** and **Battery Discharge Today** read a
permanent `unknown` on the controller device. They're inverter-level sensors whose
computed values route through givenergy-modbus's per-model battery-energy source
map, and the EMS controller (`Model.EMS_COMMERCIAL`) has no entry there — so the
fields resolve to `None`. The 0x11 controller simply doesn't populate those
registers.

This gates both sensors off EMS via `skip_if_ems`, exactly as House Consumption and
the inverter busbar Load Power already are. The existing `_reconcile_ems_entities`
path collects every `skip_if_ems` inverter sensor, so orphaned rows are removed
automatically on upgrade — no new wiring. Directly-polled inverters (AC / HYBRID /
AIO, which *are* in the source map) are untouched and keep reporting real values.

Tests: extended the exhaustive `skip_if_ems` boundary assertion, the fresh-install
"gated off EMS" test, and the upgrade-removal test in `test_ems.py`.

The other half of #221 — the "remaining battery energy" question — is a separate
discussion (it's a direct firmware register, not SOC × capacity) and stays on the
issue pending a reading from the reporter.

### DIFF
--- a/custom_components/givenergy_local/__init__.py
+++ b/custom_components/givenergy_local/__init__.py
@@ -426,10 +426,12 @@ def _retired_inverter_unique_ids(serial: str) -> set[str]:
     """Unique IDs of the entities retired on an EMS plant (#201).
 
     On an EMS controller the inverter-level *controls* are suppressed (the EMS
-    slots are authoritative) and a few *derived* inverter sensors are gated via
-    `skip_if_ems` (House Consumption), plus the dropped duplicate `ems_status`
-    aggregate. The rest of the inverter sensors stay — the 0x11 block carries real
-    plant data (PV/grid/battery/AC) — so they are deliberately NOT in this set.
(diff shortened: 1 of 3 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the current head and have no blocking concerns. The modbus model has no authoritative daily battery-energy source for the EMS controller, the topology gate leaves direct and managed inverter entities untouched, and upgrade cleanup is limited to the controller serial plus the exact gated keys. The focused EMS suite (32 passed), full suite (522 passed), Ruff, mypy, hassfest, and CI are clean.

PR #226, opened 2026-06-25
feat: raise repair when inverter clock drifts from HA

dewet22:
## Summary

Adds a fixable HA repair that fires when the inverter's RTC has drifted more than 10 minutes from Home Assistant's time. A drifted inverter clock silently breaks time-of-use charge/discharge scheduling (and Predbat) because the inverter acts on its own wall clock.

**What's new:**

- Raises a `WARNING`-severity repair, `system_time_drift_{entry_id}`, when `|inverter_clock − HA_time| ≥ 10 min`. Fires on every coordinator poll via a listener registered at entry setup; uses `ir.async_create_issue` / `async_delete_issue` (idempotent, no churn).
- **One-click Fix** writes the current HA time to the inverter RTC (`commands.set_system_date_time(dt_util.now())` — the same write as `set_system_datetime`) then triggers a refresh. The repair clears itself after the next poll confirms the clock is back in window.
- The repair description includes the opt-out instructions inline: open Configure and untick *Warn when the inverter clock drifts from Home Assistant*.
- New per-entry options toggle `warn_clock_drift` (default ON). Users who deliberately keep their inverter in UTC or another zone can switch it off; toggling it off also clears any standing repair.
- Battery-data-only entries are skipped (clock sensors aren't surfaced for those entries).

**Tests:** 14 new in `tests/test_system_time_drift.py` covering the pure helper, raise/clear/guard paths, end-to-end listener wiring at setup, fix flow (form render, clock write, disconnect guard), and the options toggle render/persist test in `test_config_flow.py`.

## Test plan
- [ ] CI green
- [ ] Manual: set inverter clock ~15 min out → repair appears naming the drift; click Fix → clock syncs, repair clears next poll
- [ ] Untick toggle → repair clears and stays away
- [ ] Battery-data-only entry: no repair raised

### DIFF
--- a/custom_components/givenergy_local/__init__.py
+++ b/custom_components/givenergy_local/__init__.py
@@ -21,7 +21,7 @@ from homeassistant.components.persistent_notification import (
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import CONF_HOST, CONF_PORT
 from homeassistant.const import __version__ as HA_VERSION
-from homeassistant.core import HomeAssistant, ServiceCall
+from homeassistant.core import HomeAssistant, ServiceCall, callback
 from homeassistant.exceptions import HomeAssistantError
 from homeassistant.helpers import config_validation as cv
 from homeassistant.helpers import device_registry as dr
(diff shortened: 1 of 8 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the drift calculation, listener and repair lifecycle, options handling, and fix path, including the modbus write-response cache update that lets the requested refresh clear the issue. I found no blocking concerns. The full local suite passes (537 tests), and CI is green.

PR #227, opened 2026-06-26
feat: expose self-consumption energy sensors (#223)

dewet22:
Closes #223.

## Summary

Adds **Self Consumption Today** and **Self Consumption Total** sensors — PV generation used on-site (`max(0, PV − grid export)`), mirroring GivTCP's `self_consumption_energy`. These complete the "self-used vs exported" split that dashboards (the FR came from setting up `pv-energy-donut-card`) previously had to derive with template helpers.

Both are thin passthrough sensors over the `e_self_consumption_today` / `e_self_consumption_total` `@computed_field`s added in **givenergy-modbus 2.5.12**; the pin is bumped to `>=2.5.12` across `manifest.json`, `pyproject.toml`, and `uv.lock`.

## Design notes

- **`monotonic=True` on both.** The library fields are differences of two cumulative counters (PV vs export), so they dip transiently on poll skew *and* genuinely decrease when the battery exports to grid. A `TOTAL_INCREASING` energy sensor that decreases is misread by HA's recorder as a meter reset, so the existing monotonic clamp (same one House Consumption Today uses) holds the high-water mark. This slightly overstates self-consumption across a battery-to-grid window — that's the documented trade-off for a monotonic Energy-dashboard total, and matches the library's stated contract for these fields.
- **`skip_if_ems=True`** — mirrors House Consumption Today. It's the same kind of derived PV/grid figure whose per-controller semantics aren't yet validated on EMS hardware. Easily revisited once confirmed on an EMS plant.
- **`skip_if_none=True`** — the fields are `SinglePhaseInverter`-only; three-phase units don't expose them, so the sensors drop rather than orphan as unavailable.

The **PV-direct-to-load** sensor mentioned as a bonus in #223 is a different derivation (`load − import − battery_discharge`) and lower priority — left for a follow-up.

## Tests

- `test_self_consumption_sensors_created` — both surface as kWh energy sensors with the expected state on single-phase.
- `test_self_consumption_sensors_absent_when_field_missing` — `skip_if_none` drops both when the field is absent.
- Extended the EMS gate tests (`test_inverter_sensors_kept_on_ems_except_gated`, and the exact gated-set pin in `test_inverter_sensors_gated_off_ems`) to cover the two new keys.
- Mock fixture gains coherent values (`9.1 = 11.2 − 2.1`, `4207.8 = 5100.2 − 892.4`).

Full suite: 539 passing.

## Test plan
- [ ] CI green
- [ ] Manual: confirm both sensors appear with kWh / energy device class and feed the donut card's consumption ring

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -76,6 +76,12 @@ class GivEnergyInverterSensorDescription(SensorEntityDescription):
     # session. Use for computed TOTAL_INCREASING sensors whose source value can
     # transiently dip due to multi-register polling skew.
     monotonic: bool = False
+    # Only meaningful with monotonic=True. True (default) runs the daily-counter
+    # clamp, which admits the genuine midnight reset back to ~0. False runs a pure
+    # high-water clamp for lifetime counters that never reset: it never re-baselines
+    # at the day boundary, so a midnight dip can't surface as a fake meter reset in
+    # HA statistics (#223).
+    monotonic_resets_daily: bool = True
     # Model field the value_fn actually reads, when it differs from `key` (the
     # key is pinned by unique_id stability). Without this, renamed direct-register
     # sensors resolve no IR source and silently miss the stale-bank availability
@@ -739,14 +745,17 @@ INVERTER_SENSORS: tuple[GivEnergyInverterSensorDescription, ...] = (
     GivEnergyInverterSensorDescription(
         # Lifetime self-consumption (max(0, pv_total - export_total), modbus 2.5.12).
         # Also a difference of cumulative counters that can transiently dip / decrease
-        # on battery-to-grid export, so it carries the monotonic clamp too — unlike
-        # the single-register e_load_total above, which needs none.
+        # on battery-to-grid export, so it carries a monotonic clamp too — but the
+        # non-daily (pure high-water) variant: a lifetime counter never resets, so the
+        # daily clamp's midnight re-baseline would expose an overnight dip as a fake
+        # meter reset in HA statistics (#223).
         key="e_self_consumption_total",
         name="Self Consumption Total",
         native_unit_of_measurement=UnitOfEnergy.KILO_WATT_HOUR,
         device_class=SensorDeviceClass.ENERGY,
         state_class=SensorStateClass.TOTAL_INCREASING,
         monotonic=True,
+        monotonic_resets_daily=False,
         value_fn=lambda inv: getattr(inv, "e_self_consumption_total", None),
         skip_if_none=True,
         skip_if_ems=True,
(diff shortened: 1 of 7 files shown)


dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -711,6 +711,46 @@ def _grid_import_power(inv: InverterModel) -> float | None:
...
+        # Lifetime self-consumption (max(0, pv_total - export_total), modbus 2.5.12).
+        # Also a difference of cumulative counters that can transiently dip / decrease
+        # on battery-to-grid export, so it carries the monotonic clamp too — unlike
+        # the single-register e_load_total above, which needs none.
+        key="e_self_consumption_total",
+        name="Self Consumption Total",
+        native_unit_of_measurement=UnitOfEnergy.KILO_WATT_HOUR,
+        device_class=SensorDeviceClass.ENERGY,
+        state_class=SensorStateClass.TOTAL_INCREASING,
+        monotonic=True,
```
dewet22-codex, line 757: [R] I think the lifetime sensor needs a different clamp from the daily counters. `monotonic=True` currently also enables the calendar-day reset path in `GivEnergyInverterSensor.native_value`: after midnight, an unchanged total marks `_monotonic_reset_pending=True`, so a later drop within the reset band, or a larger drop after a polling gap, is accepted and exposed as a decrease. That makes sense for `*_today`, but for `e_self_consumption_total` any decrease will look like a meter reset in HA statistics, exactly what this clamp is trying to avoid. I’d either keep this one as a pure high-water clamp / non-daily monotonic mode, or leave the lifetime sensor out until that distinction exists.

dewet22-codex (changes requested):
I left one blocking comment on the lifetime self-consumption sensor. The dependency bump, Today sensor, EMS gating and skip-if-missing path otherwise look consistent to me, and CI is green; the remaining issue is that the lifetime total currently uses the daily-reset-aware monotonic clamp.

custom_components/givenergy_local/sensor.py
dewet22: Good catch — you're right that `monotonic=True` was also enabling the daily-counter reset path, and the midnight re-baseline would expose an overnight dip on the lifetime total as a fake meter reset. I've gone with the first option: added a `monotonic_resets_daily` flag (default `True`, so the `*_today` clamp is unchanged) and set it `False` on `e_self_consumption_total` to select a pure high-water clamp — it holds the maximum, never re-baselines at the day boundary, and the restore path always seeds from the last persisted value. Regression test `test_lifetime_monotonic_holds_dip_and_never_resets_at_midnight` pins the cross-midnight case. Fixed in 8f21926.

dewet22-codex:
I’m starting a re-review pass on this now.

dewet22-codex (approved):
I’ve re-reviewed the latest head and the fix addresses my previous concern. `e_self_consumption_total` now uses the non-resetting high-water clamp while the existing daily-counter behaviour remains unchanged for the `*_today` sensors. The focused local sensor suite passes (`107 passed`) and CI is green.

PR #230, opened 2026-06-26
fix: suppress clock-drift repair on EMS controllers

dewet22:
## Summary

The clock-drift repair (#219 follow-up) offered a one-click fix that writes HR(35). On an EMS plant the modbus library refuses that write — it models the controller as a non-inverter peer, so HR(35) isn't in the EMS write allow-list and `one_shot_command` raises `InvalidPduState` before anything reaches the wire. The result was a 500 ("Config flow could not be loaded") when opening the repair.

Reading the clock still works on EMS, and that read — surfaced via the **System Time** datetime entity — is what Predbat consumes. So this keeps detection and drops only the local write-fix, which can't function on EMS:

- **Repair:** on EMS the drift is raised as a **non-fixable** notice (`system_time_drift_ems`) that points to the GivEnergy app/portal, rather than a Fix button. Non-EMS behaviour is unchanged (still one-click fixable).
- **System Time entity:** its `async_set_value` now raises a clean `HomeAssistantError` on EMS instead of letting the library's `InvalidPduState` surface as a traceback. The entity stays for read/visibility.

The controller normally re-syncs its clock from the cloud, so on real hardware this often clears on its own.

## Test plan

- [x] `test_ems_controller_raises_non_fixable_notice` — EMS drift raises a non-fixable `system_time_drift_ems` issue
- [x] `test_set_system_time_rejected_on_ems` — setting System Time on EMS raises `HomeAssistantError`, never reaches the client
- [x] Existing fixable-path and datetime tests unchanged
- [x] Full suite: 544 passing

### DIFF
--- a/custom_components/givenergy_local/__init__.py
+++ b/custom_components/givenergy_local/__init__.py
@@ -616,14 +616,21 @@ def _check_system_time_drift(
         "system_time": system_time.strftime("%Y-%m-%d %H:%M"),
         "ha_time": now.strftime("%Y-%m-%d %H:%M"),
     }
+    # On an EMS plant the controller's clock can't be set locally: the modbus library
+    # models the controller as a non-inverter peer and refuses HR(35) writes, so the
+    # one-click fix would error. Surface the drift as a non-fixable notice that points
+    # to the GivEnergy app instead (the controller normally re-syncs from the cloud).
+    # Reading the clock still works, so detection — what Predbat consumes via the
+    # System Time entity — is unaffected.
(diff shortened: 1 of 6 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the EMS clock-drift guard and found no blocking concerns. The detector now clears any standing drift repair and returns before raising a new issue on EMS plants, which matches the HR(35) write limitation and prevents the non-applicable fix path from being offered once the coordinator runs. The focused clock-drift tests passed locally, and CI is green.

PR #232, opened 2026-06-26
fix: use getattr for e_pv_direct_today value_fn (3-phase AttributeError)

dewet22:
## Summary

`e_pv_direct_today` is a `@computed_field` defined only on `SinglePhaseInverter`, but `Plant.inverter` is polymorphic — it returns a `ThreePhaseInverter` on three-phase plants. The #229 review swapped the value_fn from `getattr(inv, "e_pv_direct_today", None)` to direct attribute access, which shipped in v1.3.26. On a three-phase non-EMS plant the setup filter evaluates `value_fn(inverter)` eagerly, and `inv.e_pv_direct_today` raises `AttributeError`.

The platform's try/except catches it, so the sensor is still correctly skipped — nothing actually breaks — but it logs a misleading `WARNING "Skipping sensor e_pv_direct_today: its value_fn raised at setup (library field drift?)"` on every three-phase startup. The field hasn't drifted; it's simply single-phase-only.

## Fix

Revert the value_fn to `getattr(inv, "e_pv_direct_today", None)`, matching the pattern already used by `e_self_consumption_*` and `battery_nominal_*`. getattr returns `None` on models that lack the field, and `skip_if_none` then drops the sensor cleanly without the warning. (`single_phase_only=True` would also work, but the closest sibling sensors use getattr, and getattr additionally covers the AC-coupled/AIO/non-GEN1 None cases the flag wouldn't.)

## Test

Adds `test_inverter_value_fns_resolve_against_three_phase_model`, a companion to the existing single-phase resolution test — whose blind spot (never exercising `ThreePhaseInverter()`) is what let this slip through. It runs every non-`single_phase_only` value_fn against a real `ThreePhaseInverter` and asserts none raise: fails before this change, passes after. The broad form also guards against any future direct-access regression of the same class (none surfaced beyond `e_pv_direct_today`).

Full suite green at 95% coverage; mypy and ruff clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sensor handling so a missing inverter value no longer causes errors and is safely omitted when unavailable.
* **Tests**
  * Added coverage to verify inverter sensor values resolve correctly across supported inverter types, including three-phase models.

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -806,7 +806,9 @@ INVERTER_SENSORS: tuple[GivEnergyInverterSensorDescription, ...] = (
         device_class=SensorDeviceClass.ENERGY,
         state_class=SensorStateClass.TOTAL_INCREASING,
         monotonic=True,
-        value_fn=lambda inv: inv.e_pv_direct_today,
+        # getattr, not direct access: SinglePhaseInverter-only computed field, absent on
+        # the polymorphic ThreePhaseInverter — returns None there, dropped by skip_if_none.
+        value_fn=lambda inv: getattr(inv, "e_pv_direct_today", None),
         skip_if_none=True,
         skip_if_ems=True,
     ),
--- a/tests/test_sensor.py
+++ b/tests/test_sensor.py
@@ -74,6 +74,30 @@ def test_inverter_value_fns_resolve_against_real_model():
         d.value_fn(inv)
 
 
+def test_inverter_value_fns_resolve_against_three_phase_model():
+    """Companion to test_inverter_value_fns_resolve_against_real_model for the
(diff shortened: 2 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
I have reviewed feb68d0 and do not see blocking concerns. This restores the optional-field behaviour for e_pv_direct_today on the polymorphic inverter model: ThreePhaseInverter lacks the SinglePhaseInverter computed field, so returning None lets skip_if_none drop the sensor without emitting the field-drift warning.

The added ThreePhaseInverter value_fn coverage is a useful guard against this class of regression for future sensor descriptors too.

Verified locally: uv run pytest tests/test_sensor.py -k "inverter_value_fns_resolve or pv_direct_today". CI is green.

PR #239, opened 2026-06-29
feat: HV per-module battery sensors + per-cell exposure option (#179)

dewet22:
## Summary

Adds the **per-module (BMU)** layer of HV battery stacks — the remaining piece of #179 — now that `givenergy-modbus` 2.7.0 wire-confirmed the per-cell decode (against lamztib's GIV-3HY-11 in #174). The per-stack BCU sensors already shipped in #95.

Each HV module becomes its own HA device, nested under its HV-stack device, carrying **roll-up** sensors: min / max / delta cell voltage and min / max cell temperature.

## Per-cell exposure option

Individual per-cell entities (24 voltages + 24 temps per HV module — ~288 for a six-module stack — plus the existing **LV-pack** and **AIO-module** cells) are now gated behind a new global option, **"Expose per-cell battery details"**:

- **New installs:** default **off** (roll-ups only) — written explicitly at config-flow creation.
- **Existing installs:** have no option key, so it resolves to **on** — their current per-cell entities are preserved (non-breaking, no migration needed).
- Toggling it off reconciles away the orphaned per-cell rows (`_reconcile_per_cell_entities`); roll-ups and aggregates (`v_cells_sum`) are always kept.

The asymmetric default (legacy read `True`, new-entry create `False`, options-flow default `True`) is the whole migration story — there's no version bump.

## Changes
- `const.py` — `CONF_EXPOSE_PER_CELL`.
- `config_flow.py` — write `False` on create; options-flow toggle.
- `sensor.py` — split per-cell descriptors out of the LV/AIO sets; AIO + HV per-module roll-ups; `GivEnergyHvModuleSensor` (serial-keyed resolver, `via_device` → stack).
- `__init__.py` — `_reconcile_per_cell_entities`.
- `strings.json` / `translations/en.json` — option label + description.

## Test plan
- [x] `uv run pytest` — 566 passed (new: HV module creation/roll-up values, per-cell gating across LV/AIO/HV, default-on-if-absent, options round-trip, reconcile).
- [x] `uv run ruff check` + `mypy` clean.
- [ ] Manual on lamztib's HV plant: roll-ups + module devices by default; enabling the option + reload surfaces the 24+24 per-cell entities; disabling + reload removes them.

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -160,7 +160,9 @@ def _present_cells(obj: Any, prefix: str, count: int) -> list[Any]:
     Unused/unscanned cell slots read exactly 0 — for both voltages (~3.3 V when
     real) and temperatures. An exact-0 temperature is not a plausible real-world
     reading: these packs self-heat and operate above 0 °C even on the coldest
-    winter nights, so 0 is a reliable "not populated" sentinel for both types.
+    winter nights. The co-occurrence of non-zero temperatures on other cells in
+    the same module confirms the zero is a "not populated" sentinel rather than a
+    legitimate cold reading, making it safe to exclude for both types.
     Values come off the model via getattr, typed Any like other value_fns.
     """
     return [
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -155,16 +155,18 @@ def _hv_module_attr(name: str) -> Callable[[Bmu], Any]:
 
 
 def _present_cells(obj: Any, prefix: str, count: int) -> list[Any]:
-    """Non-zero, non-None cell readings `prefix_01`..`prefix_{count}` off `obj`.
+    """Non-None cell readings `prefix_01`..`prefix_{count}` off `obj`.
 
-    Unused cells in a partially-populated pack read ~0, so they're excluded from
-    the roll-ups below rather than dragging the min to zero. Values come off the
-    model via getattr, so they're typed Any like the other value_fns here.
+    Voltage cells reading 0 V mark unused slots in a partially-populated pack
+    (real cells are ~3.3 V) and are excluded so the roll-ups aren't dragged to 0.
+    Temperature cells reading 0 °C are a legitimate reading in cold conditions and
+    are kept. Values come off the model via getattr, typed Any like other value_fns.
     """
     return [
         v
         for i in range(1, count + 1)
-        if (v := getattr(obj, f"{prefix}_{i:02d}", None)) not in (None, 0)
+        if (v := getattr(obj, f"{prefix}_{i:02d}", None)) is not None
+        and not (v == 0 and prefix.startswith("v_"))
     ]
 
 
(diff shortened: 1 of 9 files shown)


dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -1802,6 +2013,34 @@ async def async_setup_entry(
             GivEnergyHvStackSensor(coordinator, description, stack_index)
             for description in HV_STACK_SENSORS
         )
+        # HV per-module (BMU) devices (#179), nested under the stack. Roll-ups are
+        # always created; the 24+24 per-cell entities are gated behind
+        # expose_per_cell. Dedup by serial like the AIO loop — a placeholder or
+        # repeated read would otherwise collide on unique_id.
+        seen_bmu_serials: set[str] = set()
```
dewet22-codex: [P2] This duplicate-serial set is recreated for each stack, but the BMU unique IDs and device identifiers are keyed only by the module serial. If two stacks decode the same placeholder/garbled serial, both stacks will try to create entities such as `HV..._cell_voltage_min`; HA then sees duplicate unique IDs, and `_bmu()` would resolve serial-keyed entities to the first matching module. I think the guard needs to be scoped across the whole HV module enumeration, declared before the stack loop, so cross-stack duplicates are skipped the same way in-stack duplicates are.

custom_components/givenergy_local/sensor.py
```diff
@@ -135,6 +140,48 @@ def _bcu_attr(name: str) -> Callable[[Bcu], Any]:
...
+
+    Unused cells in a partially-populated pack read ~0, so they're excluded from
+    the roll-ups below rather than dragging the min to zero. Values come off the
+    model via getattr, so they're typed Any like the other value_fns here.
+    """
+    return [
+        v
+        for i in range(1, count + 1)
+        if (v := getattr(obj, f"{prefix}_{i:02d}", None)) not in (None, 0)
+    ]
```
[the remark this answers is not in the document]
dewet22: Fair catch. Voltage cells reading 0 V are genuine unused slots (~3.3 V otherwise), so zero-exclusion is right there. 0 °C is a valid temperature in cold conditions, so the zero filter shouldn't have applied to temperature prefixes. Fixed by keying the zero-exclusion on `prefix.startswith("v_")`.

custom_components/givenergy_local/sensor.py
dewet22: Correct — a garbled serial appearing in more than one stack would pass the per-stack guard, then collide on unique_id and cross-wire `_bmu()` resolution to the first module that decoded that serial. Fixed by moving `seen_bmu_serials` outside the stack loop, mirroring the intent of `seen_module_serials` on the AIO side.

dewet22-codex:
I’ve started a re-review pass on this PR.

dewet22-codex (approved):
I’ve re-reviewed the latest head and do not have any blocking concerns. The HV BMU duplicate-serial guard is now scoped across the stack enumeration, which addresses the cross-stack unique-id collision case I raised, and the per-cell option/sensor paths still look consistent.

Checked locally with: uv run pytest tests/test_sensor.py tests/test_init.py::test_reconcile_per_cell_entities_removes_cells_when_disabled tests/test_config_flow.py::test_options_flow_expose_per_cell_defaults_on_for_legacy_entry tests/test_config_flow.py::test_options_flow_expose_per_cell_persists_off

PR #240, opened 2026-06-29
feat: adopt givenergy-modbus 2.8.0 + fault/warning code sensors

dewet22:
## What

Deliberate adoption of givenergy-modbus 2.8.0, pinned exactly (`==2.8.0`) in both `pyproject.toml` and `manifest.json` rather than left floating. Since both sides of this ecosystem are maintained together, syncing on each modbus release is easier to reason about than auto-floating onto an upstream patch that wasn't validated against this integration.

## Changes

- **Pin `givenergy-modbus==2.8.0`** (both pins + `uv.lock`).
- **Fault Code sensor migrated.** 2.8.0 splits the old combined `fault_code` into `inverter_fault_code` (IR39) + `inverter_warning_code` (IR40) and removes the deprecated `fault_code` alias from `ThreePhaseInverter` — direct access raised `AttributeError` on 3-phase. The sensor now reads `inverter_fault_code`; the entity key stays `fault_code`, so the entity ID, recorder history, and the dashboard row keyed on it are unchanged.
- **New Warning Code sensor** for `inverter_warning_code` (IR40), the warning half that the split newly exposes.

## Heads-up: Inverter Export Total discontinuity

2.8.0 rescales `e_inverter_export_total` to 0.1 kWh (the library's correction — the prior value read 10× high). The surfaced total drops to a tenth on the first poll after upgrading. As a `TOTAL_INCREASING` sensor this shows as a one-time downward step in the Energy Dashboard rather than a negative spike; no code change is needed, but it's worth calling out in the release note.

## Test plan

- [x] `ruff check` / `ruff format` clean
- [x] `mypy` clean
- [x] `pytest --cov` — 566 passed, 94.94% (floor 90)
- [x] `npm test` — 42 passed
- [x] `test_inverter_value_fns_resolve_against_three_phase_model` (the regression guard that failed on the `fault_code` break) now green

## Summary by CodeRabbit

* **New Features**
  * Added a separate diagnostic warning code sensor so inverter fault and warning information are shown independently.
* **Bug Fixes**
  * Updated the existing fault code sensor to report the correct inverter fault value.
* **Chores**
  * Pinned the modbus dependency to a specific supported release for consistency.
  * Aligned test data with the updated fault and warning code handling.

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -286,7 +286,21 @@ INVERTER_SENSORS: tuple[GivEnergyInverterSensorDescription, ...] = (
     GivEnergyInverterSensorDescription(
         key="fault_code",
         name="Fault Code",
-        value_fn=lambda inv: inv.fault_code,
+        # modbus 2.8.0 split the old combined fault_code into two 16-bit words and
+        # dropped the deprecated alias from ThreePhaseInverter (it raised on 3ph).
+        # Read IR39 directly; key stays "fault_code" to keep the entity ID, recorder
+        # history, and the dashboard row that's slugged on it.
+        value_fn=lambda inv: inv.inverter_fault_code,
+        entity_category=EntityCategory.DIAGNOSTIC,
+    ),
+    GivEnergyInverterSensorDescription(
+        # key is "warning_code" (not the library attr inverter_warning_code) to
+        # mirror its sibling "fault_code" — both are user-facing slugs, matching
+        # the existing convention where key="fault_code" reads inverter_fault_code.
+        key="warning_code",
+        name="Warning Code",
+        # IR40 — the warning half split out of the old fault_code in modbus 2.8.0.
+        value_fn=lambda inv: inv.inverter_warning_code,
         entity_category=EntityCategory.DIAGNOSTIC,
     ),
     GivEnergyInverterSensorDescription(
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,12 @@ version = "0.0.0"
 description = "Home Assistant custom component for GivEnergy inverters via local Modbus TCP"
 requires-python = ">=3.14.2"
 dependencies = [
-    "givenergy-modbus>=2.7.1,<3.0.0",
+    # Exact-pinned, not a floating range: we own both sides of this ecosystem,
+    # so syncing deliberately on each modbus release (and keeping the dev venv
+    # identical to the HA-runtime pin in manifest.json) beats auto-floating onto
+    # an upstream patch that wasn't validated against this integration. Bump both
+    # this and manifest.json together.
+    "givenergy-modbus==2.8.0",
 ]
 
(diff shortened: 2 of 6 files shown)


dewet22-codex:

pyproject.toml
```diff
@@ -9,7 +9,12 @@ version = "0.0.0"
 description = "Home Assistant custom component for GivEnergy inverters via local Modbus TCP"
 requires-python = ">=3.14.2"
 dependencies = [
-    "givenergy-modbus>=2.7.1,<3.0.0",
+    # Exact-pinned, not a floating range: we own both sides of this ecosystem,
+    # so syncing deliberately on each modbus release (and keeping the dev venv
+    # identical to the HA-runtime pin in manifest.json) beats auto-floating onto
+    # an upstream patch that wasn't validated against this integration. Bump both
+    # this and manifest.json together.
+    "givenergy-modbus==2.8.0",
```
dewet22-codex, line 17: [P2] This switches the dependency shape to an exact `==` pin, but `.github/workflows/bump-givenergy-modbus.yml` still only matches `"givenergy-modbus>=..."` when it updates `pyproject.toml` and `manifest.json`. The next manual modbus bump would print “no givenergy-modbus pin found” for both files, then `uv lock --upgrade-package` would have no changed requirement to solve against. I think the bump workflow needs to learn the new exact-pin shape in the same PR, otherwise the deliberate sync workflow described in this comment stops working after this adoption.

custom_components/givenergy_local/sensor.py
```diff
@@ -286,7 +286,18 @@ def _grid_import_power(inv: InverterModel) -> float | None:
...
-        value_fn=lambda inv: inv.fault_code,
+        # modbus 2.8.0 split the old combined fault_code into two 16-bit words and
+        # dropped the deprecated alias from ThreePhaseInverter (it raised on 3ph).
+        # Read IR39 directly; key stays "fault_code" to keep the entity ID, recorder
+        # history, and the dashboard row that's slugged on it.
+        value_fn=lambda inv: inv.inverter_fault_code,
+        entity_category=EntityCategory.DIAGNOSTIC,
+    ),
+    GivEnergyInverterSensorDescription(
+        key="inverter_warning_code",
```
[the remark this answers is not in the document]
dewet22: Agreed — changed to `key="warning_code"` to mirror the `fault_code` sibling. Both keys are user-facing slugs (matching the existing convention where `key="fault_code"` reads `inverter_fault_code`), so the unique_id stays consistent across the diagnostic sensors.

pyproject.toml
dewet22: Fixed in this PR. The rewrite regex only matched `>=`, so post-adoption it would have reported "no pin found" for both files and `uv lock --upgrade-package` would have had nothing to re-solve. It now matches the requirement whatever operator it carries and normalises the rewrite to `==<new>`; a dry-run against both files confirms each rewrites to `givenergy-modbus==<new>`.

PR #241, opened 2026-06-29
chore(deps): adopt givenergy-modbus 2.9.2

dewet22:
Adopts givenergy-modbus **2.9.2**, completing the modbus #268 epic. Exact pin bumped `==2.8.0` → `==2.9.2` in `pyproject.toml` + `manifest.json` (+ `uv.lock`), per the deliberate-sync policy.

## What 2.9.x brings
- **2.9.0** — `Plant.from_caches()` offline construction, a tri-state register-block presence marker, and an internal `detect()` refactor (`_strategise`/`_probe_ranges`/`_derive_capabilities`) with stale-cache eviction on every probe-failure path. No public API change.
- **2.9.1** — `refresh(max_age=)` generalises skip-if-fresh to every IR bank; the old `ir0_max_age=` is now a deprecated alias (removed in 3.0).
- **2.9.2** — `refresh()` skips banks `detect()` marked absent (~2s/tick saving in the capabilities-as-target case).

## Why no code change
The coordinator's entire modbus call surface is `detect(prior=, probe_timeout=, probe_retries=)`, `refresh(retries=)` and `load_config(retries=)` — all keep identical signatures in 2.9.2 (verified by introspection). We never passed `ir0_max_age`, so the deprecation doesn't apply and `deprecation watch` stays green. Adopting `max_age=` is deliberately deferred to a separate, independently-validated change.

## Why rc
The headline change is behavioural: the `detect()` refactor reworks the detection hot path that surfaces batteries / EMS / HV stacks — the same machinery behind the #233 enumeration deadlock. So this is being cut as a **v1.3.32rc** pre-release for real-hardware validation before a stable v1.3.32.

## Test plan
- [x] `ruff check` / `ruff format --check` clean
- [x] `mypy` clean
- [x] 566 Python tests green against 2.9.2
- [x] 42 JS tests green
- [ ] Real-hardware validation via HACS pre-release (detection, battery/EMS/HV enumeration, poll stability)

### DIFF
--- a/custom_components/givenergy_local/manifest.json
+++ b/custom_components/givenergy_local/manifest.json
@@ -13,7 +13,7 @@
   "iot_class": "local_polling",
   "issue_tracker": "https://github.com/dewet22/givenergy-hass/issues",
   "requirements": [
-    "givenergy-modbus==2.9.2"
+    "givenergy-modbus==2.9.3"
   ],
   "version": "1.1.0"
 }
--- a/custom_components/givenergy_local/manifest.json
+++ b/custom_components/givenergy_local/manifest.json
@@ -13,7 +13,7 @@
   "iot_class": "local_polling",
   "issue_tracker": "https://github.com/dewet22/givenergy-hass/issues",
   "requirements": [
-    "givenergy-modbus==2.8.0"
+    "givenergy-modbus==2.9.2"
   ],
   "version": "1.1.0"
 }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ dependencies = [
     # identical to the HA-runtime pin in manifest.json) beats auto-floating onto
     # an upstream patch that wasn't validated against this integration. Bump both
     # this and manifest.json together.
-    "givenergy-modbus==2.9.2",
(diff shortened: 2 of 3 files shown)


dewet22:
Holding this — the rc surfaced a regression on hardware. Running v1.3.32rc1 overnight, the primary LV battery dropped out permanently after the inverter's ~02:00 cloud-sync reconnect: the splice-guard baseline gets seeded from a partially-spliced frame at the reconnect (zero cell/temp registers, but correct firmware and capacity), and every healthy read afterwards is rejected against it, with no recovery.

Root cause is library-side — filed as dewet22/givenergy-modbus#350 — so this is fundamentally a modbus fix, and the pin will move to whatever version carries it before this merges.

Not merging yet even though the fix is upstream: the poisoning plays out during the post-reconnect refresh cycle, so it's worth confirming nothing in how the coordinator drives `refresh()` (the cadence, the `load_config`/`refresh` interleave, the deferred `max_age` adoption) contributes as the modbus analysis develops. Keeping the rc and this PR open until that's clear and a fixed modbus build re-validates overnight.

PR #254, opened 2026-07-02
fix: options flow can't save with the experimental section (#251)

dewet22:
Fixes #251.

**Symptom:** opening **GivEnergy Local → Configure** and pressing Submit fails with *"Not all required fields are filled in"*, even though every visible field is set (see the reporter's screenshot). The options dialog is unsaveable.

**Cause:** the collapsed *Experimental features* section holds a `vol.Required` boolean. HA doesn't render a collapsed section's inner form, so that required field is never initialised in the frontend's data model; the dialog's submit-time required-fields check then walks the section, finds it undefined, and blocks. This affects **every** entry that has never opened+saved the section (none carry an `experimental` options key until then) — latent since the section shipped in v1.3.22, surfacing only now because opening options is rare.

**Fix:** inner section fields become `vol.Optional` (the documented HA pattern). The default still applies server-side and `resolve_experimental_client_kwargs` already reads via `.get(conf_key, feature.default)`, so an omitted toggle round-trips unchanged — behaviour identical, the frontend just stops treating a collapsed toggle as unfilled. The now-inaccurate comment (which asserted collapsed sections keep their inputs in the DOM — the very thing this disproves) is corrected.

**Tests:** serialises the schema as the frontend receives it and asserts the inner field is optional (verified to fail under `vol.Required`); plus an end-to-end save that omits the collapsed section. 581 Python + 42 JS green.

### DIFF
--- a/custom_components/givenergy_local/config_flow.py
+++ b/custom_components/givenergy_local/config_flow.py
@@ -187,20 +187,24 @@ class GivEnergyLocalOptionsFlow(OptionsFlow):
         # Surface the collapsed "Experimental features" group only once at least one
         # flag exists, so the header never appears empty (the registry ships empty).
         if EXPERIMENTAL_FEATURES:
-            # Seed schema defaults from the currently-saved values so that a
-            # collapsed or omitted section round-trips correctly.  This works
-            # because HA's frontend either (a) submits the rendered (pre-filled)
-            # values for a collapsed section, in which case the submitted value
-            # wins, or (b) omits the section key entirely, in which case the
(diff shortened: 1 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed this and have no blocking concerns. The schema change is scoped to the nested experimental section, the existing/omitted/explicit-false paths are covered, and the options-flow preservation invariant held in the focused run. Local check: uv run pytest tests/test_config_flow.py (25 passed). CI is green on the reviewed head.

PR #256, opened 2026-07-02
feat!: remove passive mode (#253)

dewet22:
Closes #253.

**Why remove rather than fix:** passive mode's freshness check (fail after ~60 s without a cache update) could never coexist with the GivEnergy cloud's ~5-minute polling cadence — the setup users most plausibly ran it with. The result was a spurious refresh failure on ~87% of ticks and entities flapping unavailable while everything was actually working; the reporter's screenshot shows the counter climbing at exactly one per 30 s tick for 15 hours. The contention fear that motivated passive mode hasn't materialised on any hardware we've encountered — multiple active clients at 30-second intervals are well within what the dongle handles.

**What changes:** the coordinator always polls actively (`_passive_update`, `_check_cache_freshness` and their state removed); the setup/reconfigure field, constants and translation labels go; README loses the passive section, gains a removal note, and the troubleshooting entries are updated. A legacy `passive` key left in entry data is deliberately inert — entries that had it enabled simply start polling actively, which is the fix. No migration needed.

**Tests:** 9 passive tests removed; new test pins the legacy-inert-key behaviour (`passive: true` entry sets up and issues its own load_config+refresh). 573 Python + 42 JS green.

## Summary by CodeRabbit

* **New Features**
  * The integration now always uses active polling, including for older setups that previously relied on passive mode.
  * Concurrent polling with other local clients is now supported more reliably.

* **Documentation**
  * Updated setup and troubleshooting guidance to remove references to passive mode.
  * Clarified that existing saved passive settings are ignored and normal polling continues.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -13,7 +13,7 @@ givenergy-modbus as its core library.
 - `coordinator.py` — DataUpdateCoordinator; centralises all Modbus polling via
   `client.refresh()` + `client.load_config()` (full refresh every N ticks)
 - `sensor.py`, `number.py`, `select.py`, `switch.py`, `time.py` — entity platforms
-- `config_flow.py` — UI setup (IP, port, scan interval, battery count, passive mode)
+- `config_flow.py` — UI setup (IP, port, scan interval)
 - `dashboard.py` — `generate_dashboard` service (deprecated; scheduled for removal); produces a Lovelace YAML file
 - `www/` — bundled frontend assets (e.g. `ge-cell-heatmap.js`) served at `/local/`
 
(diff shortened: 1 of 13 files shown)


dewet22-codex:

dewet22-codex (approved):
I have reviewed this and have no blocking concerns. The active polling path now ignores legacy passive entry data, the setup and reconfigure schema no longer exposes the field, and the legacy passive-entry behaviour is covered. Local checks passed: uv run pytest tests/test_init.py tests/test_config_flow.py tests/test_coordinator.py (134 passed) and uv run pytest (573 passed). GitHub Actions checks were green when I checked; CodeRabbit was still processing its review. One small housekeeping follow-up I noticed: AGENTS.md still describes config_flow.py as including passive mode and says passive mode exists. I have not treated that as blocking because the runtime path and user-facing docs are consistent in this PR.

PR #258, opened 2026-07-03
feat: Gateway sensor surface — whole-house + per-AIO telemetry (#194)

dewet22:
The follow-through on #95/#194 phase 2: a Gateway entry now surfaces real data instead of the 0/Unknown wall.

**Why everything read zero:** on a Gateway plant `plant.inverter` is a spurious all-zeros SinglePhaseInverter decoded from the gateway's own cache; the real data lives in `plant.gateway` (GatewayV1/V2). New `GATEWAY_SENSORS` + `GivEnergyGatewaySensor` (the EMS pattern) read it on the same 0x11 device: whole-house load/PV power, today's energies (load, PV, grid import/export, battery charge/discharge), voltages, per-AIO SOC/power/energies, parallel status, fault codes. Unpopulated AIO slots are gated on their empty serial.

**Held back pending upstream decode verdicts** (from the first live data): lifetime `*_total` counters (mis-scaled ~10⁴), the combined `e_aio_charge/discharge_today` pair (≠ per-AIO sum), `battery_type`.

**Suppression:** the standard inverter set is suppressed on Gateway, all control platforms + Restart gate off Gateway entries (no validated write surface), and a reconciler removes the stale rows a v1.3.38 Gateway entry created.

**Test fixture uses AberDino's real daylight values** (p_load 1127 W, per-AIO SOC 94/96, charge todays 9.9+9.3=19.2). 578 Python + 42 JS green; ruff/mypy clean.

## Summary by CodeRabbit

* **New Features**
  * Added support for Gateway devices with dedicated telemetry, including whole-house power, daily energy, and per-AIO battery readings where available.

* **Bug Fixes**
  * Prevented standard inverter sensors, switches, numbers, selects, time controls, and restart actions from appearing on Gateway setups.
  * Improved handling of existing setups so outdated entities are removed when a Gateway is detected.

* **Documentation**
  * Updated the supported devices and entity documentation to reflect Gateway support and its current limitations.

### DIFF
--- a/custom_components/givenergy_local/sensor.py
+++ b/custom_components/givenergy_local/sensor.py
@@ -196,6 +196,49 @@ class GivEnergyEmsSensorDescription(SensorEntityDescription):
     value_fn: Callable[[Ems], Any] = field(default=lambda _: None)
 
 
+@dataclass(frozen=True, kw_only=True)
+class GivEnergyGatewaySensorDescription(SensorEntityDescription):
+    """Sensor over `plant.gateway` (GatewayV1/V2 — identical field names).
+
+    On a Gateway plant `plant.inverter` is a spurious SinglePhaseInverter decoded
+    from the gateway's own 0x11 cache (all zeros), so the standard inverter sensor
+    set is suppressed and this set reads the real whole-house data from the
+    gateway register banks instead (#194).
+    """
+
+    value_fn: Callable[[Any], Any] = field(default=lambda _: None)
+    # Skip creation when the value reads None at setup — used to gate the per-AIO
+    # slots (the register map always carries three; unpopulated slots have an
+    # empty serial and their fields read as absent).
+    skip_if_none: bool = False
+
+
+def _gateway_attr(name: str) -> Callable[[Any], Any]:
+    """Return a value_fn reading `name` off the gateway model."""
+    return lambda gateway: getattr(gateway, name, None)
+
+
+def _gateway_aio_attr(slot: int, name: str) -> Callable[[Any], Any]:
+    """Read a per-AIO field, gated on AIO slot `slot` being populated.
+
+    The Gateway register map carries three AIO slots regardless of how many
+    units are installed; unpopulated slots report zeros, which must surface as
+    absent rather than as a phantom third AIO. The gate is the parallel AIO
+    count register, NOT the per-slot serial: on the live GAAA0014 shape the
+    GatewayV2 serial layout decodes aio2_serial_number as None under the
+    pinned modbus release, which would wrongly drop a real second AIO
+    (Codex catch on #258).
+    """
+
+    def value(gateway: Any) -> Any:
+        count = getattr(gateway, "parallel_aio_num", None)
+        if count is None or slot > count:
+            return None
+        return getattr(gateway, name, None)
+
+    return value
+
+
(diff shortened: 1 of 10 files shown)


dewet22-codex:

custom_components/givenergy_local/sensor.py
```diff
@@ -196,6 +196,44 @@ class GivEnergyEmsSensorDescription(SensorEntityDescription):
...
+def _gateway_aio_attr(slot: int, name: str) -> Callable[[Any], Any]:
+    """Read a per-AIO field, gated on AIO slot `slot` being populated.
+
+    The Gateway register map carries three AIO slots regardless of how many
+    units are installed; an unpopulated slot has an empty serial and reports
+    zeros, which must surface as absent rather than as a phantom third AIO.
+    """
+
+    def value(gateway: Any) -> Any:
+        if not getattr(gateway, f"aio{slot}_serial_number", None):
```
dewet22-codex: RAG: Red. I think this gate is depending on a serial field that is not reliable in the pinned modbus release yet. With givenergy-modbus 2.9.6, the live GAAA0014 Gateway shape selects GatewayV2 and decodes aio2_serial_number as None, so all AIO 2 SOC/power/energy sensors are skipped even though parallel_aio_num is 2. The synthetic test gateway sets both serials, so it does not catch the real two-AIO case this PR is adding. I would tie this to a known-good slot indicator such as the parallel AIO count, or hold this PR for a modbus release that decodes the live serial layout correctly.

dewet22-codex (changes requested):
I found one blocking issue in the per-AIO gating. With the pinned givenergy-modbus 2.9.6 and the live GAAA0014 register shape from the Gateway work, select_gateway picks GatewayV2 and aio2_serial_number decodes as None. That means this PR would omit all AIO 2 sensors on the two-AIO Gateway it is intended to support, while the synthetic fixture still passes because it sets both serials manually. I left the detail inline.

I also checked the clock/control surface while reviewing. The real all-zero inverter decode gives system_time=None, so I have not treated datetime as a separate blocker here, though the Gateway test fixture is less representative than the real model on that point.

Focused checks passed locally: uv run pytest tests/test_sensor.py; uv run pytest tests/test_button.py tests/test_number.py tests/test_select.py tests/test_switch.py tests/test_time.py tests/test_datetime.py.

custom_components/givenergy_local/sensor.py
dewet22-claude: Fixed in e92a6dd — the gate now keys on parallel_aio_num (slot > count ⇒ absent) with no serial dependency, and the test fixture sets aio2_serial_number=None to mirror the live 2.9.6 decode, so the two-AIO case is pinned against exactly this regression.

dewet22-codex:
I have started a re-review pass on this PR, focused on the previous Gateway per-AIO gating blocker.

dewet22-codex (approved):
I have re-reviewed the latest head e92a6dd against the previous Gateway per-AIO blocker. The gate now uses parallel_aio_num rather than the decoded serial fields, and the fixture keeps aio2_serial_number=None while asserting that AIO 2 sensors are still created, so the regression I was concerned about is covered.

Focused check passed locally: uv run pytest tests/test_sensor.py.

PR #259, opened 2026-07-07
feat: adopt givenergy-modbus 2.9.7 — Gateway lifetime totals + AC-side energy set

dewet22:
Follow-through on #258 once the modbus#360 fix shipped (2.9.7): `select_gateway()` defaults to V1 word order (the version-based selector was falsified by the first live unit) and the per-AIO serial layout is corrected.

**Lifts the #258 hold:** site lifetime totals (grid import/export, PV, load, battery charge/discharge), per-AIO lifetime totals (`parallel_aio_num`-gated), and the AC-side site set (`e_aio_charge/discharge` today+total) named **AIO AC …** — distinct real semantics from the battery-DC set (conversion losses between them; includes EV-charger energy on equipped sites, per AberDino's observation — this explains the 24.0 vs 19.2 discrepancy flagged in #258).

README: Gateway section updated; supported-hardware list aligned with the library's validated matrix.

Test fixture carries the V1-order sane totals from the decode verdict (grid import 12,710.8 kWh etc.). 578 Python + 42 JS green; ruff/mypy clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,12 +27,14 @@ Confirmed working on real hardware:
 - AC-coupled (Gen 1)
 - All-in-One (AIO) — including per-module battery devices with per-cell data; needs v1.2.0 or later
 - EMS controller — validated on a live two-inverter EMS plant; some polling rough edges on busy shared buses are still being smoothed out in the library
-- Gateway (V1 / V2) — validated on a Gen 1 Gateway with two AIOs; exposes whole-house load/PV power and energy, grid import/export, and per-AIO SOC and charge/discharge. Lifetime totals are held back pending a decode fix, and no controls are offered through a Gateway yet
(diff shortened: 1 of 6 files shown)


dewet22-codex:

dewet22-codex (approved):
I have reviewed the latest head ff1e44b and have no blocking concerns.

The HASS side lines up with the givenergy-modbus 2.9.7 Gateway model: the new lifetime and AC-side energy fields exist on both Gateway variants, select_gateway now returns the V1 layout used by the live GAAA0014 capture, and the per-AIO totals still respect the parallel_aio_num gate from #258.

Focused checks passed locally: uv lock --check; uv run pytest tests/test_sensor.py. I also did a small value-function sweep over the newly added Gateway sensor keys to catch missing-field or duplicate-key mistakes.

PR #262, opened 2026-07-07
feat: allow UI deletion of stale devices

dewet22:
Implements `async_remove_config_entry_device`: deletion is allowed iff none of the device's identifiers appear in the current topology (`_live_device_identifiers` mirrors every DeviceInfo shape: plant serial, pack/module/BMU serials, `{serial}_hvstack_{addr}`, `{serial}_managed`). Live devices are refused — they'd recreate on reload anyway. Unloaded entry ⇒ allowed (nothing authoritative to defend). Previously stale devices required deleting and re-adding the entire entry. 582 Python + 42 JS green.

### DIFF
--- a/custom_components/givenergy_local/__init__.py
+++ b/custom_components/givenergy_local/__init__.py
@@ -12,7 +12,7 @@ from typing import TypedDict
 import voluptuous as vol
 from givenergy_modbus.client import commands
 from givenergy_modbus.model.inverter import Model
-from givenergy_modbus.model.plant import PlantCapabilities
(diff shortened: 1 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
Reviewed this and I have no blocking concerns. The deletion hook is conservative: it only permits removal once an identifier is absent from the current coordinator topology, and the live identifiers match the DeviceInfo shapes I checked for inverter/controller/gateway, LV/AIO/HV batteries, HV stacks and EMS managed inverters. Focused check passed locally: uv run pytest tests/test_init.py.

PR #264, opened 2026-07-07
fix: drop the AIO prefix from the Gateway AC-side energy names

dewet22:
AberDino's v1.3.39 review point (#95): the AC-side set covers more than the AIOs (includes EV-charger energy on his site), so "AIO AC Charge/Discharge" misleads — plain **AC Charge/Discharge Today/Total** it is. Display names only; keys/unique_ids keep the library field names. Shipped hours ago = cheapest possible rename window. 584 Python + 42 JS green.


dewet22-codex:

dewet22-codex (approved):
Reviewed this and I have no blocking concerns. The change is limited to Gateway AC-side display names and matching README wording; the e_aio_* keys/unique_ids remain stable, so existing registry rows should keep their history while the labels become less misleading. Focused local check passed: uv run pytest tests/test_sensor.py -k gateway.

PR #266, opened 2026-07-08
fix: Gateway sensor signs, Grid Power identity, shared-key reconciler bug

dewet22:
From AberDino's line-by-line v1.3.40 comparison against GivTCP (#95):

1. **AIO Power Total + Liberty Power sign-inverted** — negated to the house convention (positive = discharging), live-verified. Liberty's semantics stay unpinned upstream; negation is empirical.
2. **p_ac1 is Grid Power** (live-matched): renamed, signed positive-export hidden-by-default per the #151 pattern, plus visible **Grid Power Import/Export** split (`p_ac1_import/export` — keys deliberately distinct from the inverter split keys, see below).
3. **Latent reconciler bug**: five keys are shared between INVERTER_SENSORS and GATEWAY_SENSORS, and `_retired_on_gateway_unique_ids` retired all inverter keys — deleting those five live gateway rows on EVERY reload since v1.3.39. Retired set now excludes gateway-reused keys; regression test pins survival.

585 Python + 42 JS green.

## Summary by CodeRabbit

* **New Features**
  * Added separate grid power import and export sensors for clearer energy tracking.
  * Improved gateway power readings so key values are shown with the correct sign.

* **Bug Fixes**
  * Prevented shared gateway sensor entries from being removed during reloads or upgrades.
  * Kept existing live gateway entities intact when reconciling sensor rows.

### DIFF
--- a/custom_components/givenergy_local/__init__.py
+++ b/custom_components/givenergy_local/__init__.py
@@ -496,10 +496,15 @@ def _retired_on_gateway_unique_ids(serial: str) -> set[str]:
     GATEWAY_SENSORS set replaces them on the same device; coordinator sensors
     use different keys and stay.
     """
-    from .sensor import INVERTER_SENSORS
+    from .sensor import GATEWAY_SENSORS, INVERTER_SENSORS
 
     keys = _inverter_control_keys()
-    keys.update(d.key for d in INVERTER_SENSORS)
+    # Five keys are shared between the sets (p_pv and the pv/load/battery
+    # lifetime totals): the gateway sensors deliberately reuse them for
+    # continuity, so retiring every inverter key would delete those live
+    # gateway rows on EVERY reload (shipped briefly in v1.3.39/40).
+    gateway_keys = {d.key for d in GATEWAY_SENSORS}
+    keys.update(d.key for d in INVERTER_SENSORS if d.key not in gateway_keys)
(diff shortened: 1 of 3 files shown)


dewet22-codex:

dewet22-codex (approved):
Reviewed a9a12ac. I don’t see any blocking concerns: the shared-key gateway reconciler now preserves the reused live gateway sensor rows, the p_ac1 split keeps distinct gateway keys, and the sign flips/import-export helpers match the stated house convention. Focused local check passed: `uv run pytest tests/test_sensor.py -k gateway` (7 passed).

dewet22-codex:
I have started a re-review pass on this PR.

dewet22-codex (approved):
Reviewed ecbeb97. I do not see any blocking concerns in the amended scope: the shared-key reconciler still preserves gateway-owned reused rows, p_ac1 has the hidden signed Grid Power row plus visible import/export split, and the AIO/Liberty values now deliberately pass through raw with the future sign correction left to the library model.

Focused local checks passed: uv run pytest tests/test_sensor.py -k gateway (7 passed), plus git diff --check.

One non-blocking housekeeping note: the PR body still says the AIO/Liberty fields are negated to the house convention, which no longer matches this head. I would align that before merge so the release context stays clear.

PR #277, opened 2026-07-09
feat: retire the Battery Discharge This Year sensor (#275)

dewet22:
Closes Ask 2 of #275 (Nick / njhcarroll).

**Battery Discharge This Year** (`e_discharge_year`) carried no useful signal on any hardware: a static meaningless figure on AIO, a flat 0 on the inverters, and unavailable on EMS. Removed:

- Drop the sensor description.
- **Reconcile the orphaned rows** on upgrade (`_remove_retired_battery_discharge_year`, runs pre-platform for all models) — so it's genuinely gone rather than left as a dead, unavailable entity, which is what Nick was seeing.
- Drop its row from the bundled dashboard strategy (`_STRATEGY_VERSION` → 13).
- Sweep the test/mock references.

Regression test pins the reconciler; 590 Python + 42 JS green, ruff + mypy clean. Intended to ride into v1.4.0. (Ask 1 — EMS Battery Charge/Discharge Today — is a separate, properly-soaked v1.4.1 feature.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,7 +298,6 @@ Some energy sensors are model-dependent: on AC-coupled and All-in-One systems th
 | Export Power Rate Limit | % | Three-phase only — installer-set grid export power-rate cap (% of rated power) |
 | Inverter Export Total | kWh | Cumulative inverter export to grid |
 | Charge from Grid Total | kWh | Cumulative grid-sourced battery charging |
-| Battery Discharge This Year | kWh | |
 | Backup Power | W | EPS port output |
 | Combined Generation Power | W | Solar + battery combined |
 | Work Time Total | h | |
--- a/custom_components/givenergy_local/__init__.py
+++ b/custom_components/givenergy_local/__init__.py
@@ -71,6 +71,7 @@ from .migrations import (
     _reconcile_gateway_entities,
     _reconcile_per_cell_entities,
     _reconcile_readability_gated_controls,
+    _remove_retired_battery_discharge_year,
 )
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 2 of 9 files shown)


dewet22-codex:
Starting a review now. I’ll focus on behaviour, Home Assistant integration invariants, and any modbus handoff assumptions in the changed paths.

dewet22-codex (changes requested):
Thanks, the runtime side looks sound: the retired entity is removed before platform setup, the sensor description and dashboard row are gone, and the focused migration/sensor/strategy tests pass locally.

One remaining public-facing reference still needs sweeping before I’d call the retirement complete: `README.md:301` still lists “Battery Discharge This Year” in the diagnostic sensor table. Since the PR removes the entity and reconciles old registry rows, keeping it in the supported-sensors table would leave the docs pointing users at a sensor that no longer exists.

Local checks run: `uv run pytest tests/test_init.py::test_upgrade_removes_retired_battery_discharge_year tests/test_sensor.py::test_aio_inapplicable_inverter_sensors_gated_on_aio tests/test_sensor.py::test_expected_sensor_count tests/test_sensor.py::test_battery_data_only_keeps_battery_drops_inverter_sensors tests/test_strategy_manifest.py` (6 passed); `git diff --check`.

dewet22-codex:
Starting a re-review now. I’ll focus on the outstanding README sweep and any new changes in that scope.

dewet22-codex (approved):
I’ve re-reviewed the updated head and have no blocking concerns now.

The outstanding README reference has been removed, and the remaining `e_discharge_year` references are only the registry cleanup and regression tests. CI is green on the updated head; locally I rechecked the stale-reference sweep and `git diff --check`.

PR #279, opened 2026-07-10
feat: add EMS battery charge/discharge energy sensors (#275)

dewet22:
Adds EMS Battery Charge/Discharge energy sensors — the second half of the v1.4.1 cycle (Nick's [#275](https://github.com/dewet22/givenergy-hass/issues/275) Ask a).

## Why

The HA Energy dashboard's battery section wants two figures: energy *into* the battery and energy *out*. On an EMS plant there's no register-backed battery-energy counter, and the managed-inverter summaries are blinded — so, exactly like the existing EMS Calculated Load Energy, these are derived by integrating a live power signal client-side. Here that's `ems.total_battery_power` (IR2090), sign-split into charge and discharge.

## What

- **Generalises `GivEnergyEmsLoadEnergySensor`** with two optional params, `source_field` (default `calc_load_power`) and `direction` (`None` | charge | discharge). The two existing load-energy sensors pass the defaults and are byte-for-byte unchanged. All the hardened integral behaviour — idempotent-per-refresh, gap-cap, `RestoreEntity` persistence, missing-sample boundary-consume, midnight reset — is inherited verbatim; the only new logic is a directional clamp (`_directional_power`).
- **Four sensors on the EMS controller device:** Battery Charge / Discharge, each Today + Total (`TOTAL_INCREASING` kWh). Nick asked for Today; Total is added to match the existing load-energy pair and give the Energy dashboard the lifetime `TOTAL_INCREASING` primitive it buckets itself.

## Sign convention — confirmed, not assumed

positive = discharge, negative = charge. Confirmed from a golden EMS capture two independent ways: the busbar power-balance closes to the watt only under this sign (battery +2511 W = load 677 W + export 1834 W), and IR2090 equals the exact sum of the managed inverters' battery power (inheriting their house convention). Isolated to a single `_directional_power` method regardless.

## Estimate caveats (inherited, documented on the class)

Client-side integral: undercounts across gaps/restarts, never overcounts; survives HA restarts.

## Tests & docs

Three new tests (`test_ems.py`): discharge accumulates on positive power / charge on negative / each reads zero on the off-sign, plus all four surface on the EMS device. The shared integral machinery is already covered by the load-energy tests. README EMS section updated.

## Soak

v1.4.1rc with Nick validating magnitude on his EMS plant (I have no EMS hardware). The sign is confirmed from a single exact power-balance tick rather than a multi-sample SOC ramp — modbus will re-confirm if a future capture catches the EMS battery moving its SOC measurably, but the balance is unambiguous.

## Verification

`ruff` clean, `mypy` clean, **595** Python tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,13 +364,14 @@ All-in-One systems expose each removable battery module separately, so on AIO ha
 
 ### EMS controller device
 
-On an EMS plant the controller device carries plant-level aggregates the inverter registers don't: managed-inverter count, calculated/measured load power, grid meter power, total battery power and remaining battery energy. Two derived energy sensors build on the load aggregate:
+On an EMS plant the controller device carries plant-level aggregates the inverter registers don't: managed-inverter count, calculated/measured load power, grid meter power, total battery power and remaining battery energy. Since the EMS rollup carries no energy counters, several energy sensors are derived by integrating those live power signals client-side:
 
 | Entity | Unit | Notes |
 |---|---|---|
 | EMS Calculated Load Energy Today / Total | kWh | Integrated client-side from *EMS Calculated Load Power* — the EMS rollup carries no energy counters, so no register-backed figure exists. An estimate, not a meter reading: it undercounts across restarts and outages (a sampling gap contributes at most one capped slice), though the accumulated value does survive Home Assistant restarts. |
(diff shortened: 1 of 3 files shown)


dewet22-codex:
Starting a review now. I’ll focus on behavioural correctness, Home Assistant lifecycle/state handling, and consistency with the surrounding integration patterns.

dewet22-codex (approved):
I’ve reviewed this and have no blocking concerns. The EMS battery charge/discharge sensors reuse the existing RestoreEntity integral path without changing the load-energy defaults, the sign split is isolated in `_directional_power`, and the four new entities are created only on EMS controller setups with stable unique IDs.

Verified with `uv run pytest tests/test_ems.py -q` and `uv lock --check`. CI is also green on the current head.

PR #282, opened 2026-07-10
feat: gate reconnect on a liveness probe (adopt modbus 2.12.0)

dewet22:
Adopts givenergy-modbus 2.12.0 and uses its new `probe_alive()` to make reconnection cheap and low-impact on a hung dongle. Closes the build half of #280 (Lever 1).

## The problem

On a marginal dongle (diagnosed from AberDino's Gen-1 Gateway capture in #95), the TCP connection drops (`connection lost (reader at EOF)`) and the dongle then refuses even its identity read for a minute or two — a Modbus-level hang. Previously each reconnect tick ran a **full detect** against the dead socket, spending ~8s (4×2s) hammering it before giving up, then retrying ~40s later.

## The fix

On a *reconnect* (`self.data is not None`), gate the heavy detect sweep behind modbus 2.12.0's `probe_alive(retries=1)` — a single cheap HR0 identity read:

- **Not alive** → `probe_alive` closes the socket and returns `False`; we raise a transient timeout, so the existing teardown + serve-last-known path runs and the next tick reprobes. No detect sweep against a dead connection.
- **Alive** → the socket is left open and the existing (robust, default-retries) `detect` runs on it to re-establish topology.

The first-ever setup (`self.data is None`) skips the probe and runs the full detect to establish topology from scratch. `retries=1` tolerates a single dropped identity frame on an otherwise-healthy reconnect (fails in ~4s on a genuine hang vs ~2s at 0) — apt for exactly the lossy dongles this targets.

This is the split modbus recommended (PR #398): a fused `detect(retries=0)` would have made the *recovery* detect fragile too, since the sweep would inherit the fail-fast retries. `probe_alive` keeps the liveness check cheap and the recovery robust.

## Scope

Pin `2.11.2 → 2.12.0`. This is Lever 1 only; deliberately **not** bundled: the exponential-backoff experiment (Lever 2, still unproven — #280), the #106 `Plant.devices` walk migration (at-my-pace), and the #399 connection-lifecycle review that also ship in 2.12.0.

## Tests

Three coordinator tests: reconnect + probe `False` serves last-known and skips detect (client torn down, one failure counted); reconnect + probe `True` runs the robust detect and polls; first-connect skips the probe entirely. The conftest mock Client gains `probe_alive` (default alive).

## Verification

`ruff` clean, `mypy` clean, **598** Python tests. Clean additive 2.12.0 adoption (no breakage).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reconnect handling by checking connection health before performing a full system scan.
  * Failed connections are now discarded promptly, allowing the next update to establish a clean connection.
  * Preserved the most recent data when a reconnect attempt temporarily fails.
  * Initial setup continues to perform the full system discovery process as expected.

### DIFF
--- a/custom_components/givenergy_local/coordinator.py
+++ b/custom_components/givenergy_local/coordinator.py
@@ -86,6 +86,13 @@ def _summarize_failures(failures: Iterable[ReadFailure]) -> str:
 PROBE_TIMEOUT_SECONDS = 1.0
 PROBE_RETRIES = 3
 
+# On a reconnect we gate the heavy detect sweep behind a cheap HR0 liveness probe
+# (givenergy-modbus 2.12.0 `probe_alive`). retries=1 tolerates a single dropped
+# identity frame on an otherwise-healthy reconnect (fails in ~4s on a genuine
+# hang vs ~2s at 0) — the lossy-dongle case (#280) is exactly where a healthy
+# reconnect can still drop a stray frame, so we don't want to bail on the first.
+RECONNECT_PROBE_RETRIES = 1
+
 # When detect() reports a DEVICE LOSS (a previously-known battery/meter/stack
 # stopped answering), re-probe a few times before believing it — a slow BMS
 # often answers on the next sweep. Kept small: this blocks _connect(), which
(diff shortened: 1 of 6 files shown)


dewet22-codex:
Starting a review now. I’ll focus on behavioural correctness, Home Assistant integration invariants, and compatibility with the modbus/client contracts.

dewet22-codex (approved):
I have reviewed this and have no blocking concerns. The reconnect-only probe gate preserves cold-start detection, skips the heavy detect sweep when the dongle is not answering, and still serves last-known data through the existing timeout-tolerance path. The 2.12.0 pin is consistent across manifest, pyproject and lockfile.

Local checks passed: uv run pytest tests/test_coordinator.py -q, uv run pytest tests/test_init.py -q, and uv lock --check.

PR #285, opened 2026-07-12
fix: keep DC battery-rate controls on directly-polled All-in-One (#281)

dewet22:
Closes #281 (the control-surface half of it).

## Background

On a directly-polled Gen-1 All-in-One, @crg-n found that the integration was hiding the battery-rate controls Predbat actually needs. His field evidence on #281: both GivTCP and GivEnergy's own control path write **HR111** when setting the charge rate, leaving HR313 untouched. So on an AIO the DC pair (HR111/112) is the operative battery-rate control, and the AC-config pair (HR313/314) is additional surface — pause/EPS/AC-side extras — not a replacement. The diagnosis here is his; this PR is the implementation.

## The defect

`number.py` gated *both* the AC-pair creation and the DC-pair suppression on `has_ac_config_block`. That capability is a register-surface fact — which models answer HR300-359 — and it's true for `{AC, AC_3PH, ALL_IN_ONE}`. The AIO carries the block but is DC-battery-backed, so gating suppression on it stripped the AIO's real controls and left only the HR313/314 pair.

## The change

Split the one flag along the two capabilities it actually spans:

- **Create the AC pair (HR313/314)** — stays on `has_ac_config_block` (register surface). The AIO keeps its AC-side extras.
- **Suppress the DC pair (HR111/112)** — moves to `is_ac_coupled` (`{AC, AC_3PH}` — control routing, "no operative DC battery"). Only genuinely AC-coupled plants lose it; the AIO keeps both pairs.

`_reconcile_ac_coupled_dc_limits` in `migrations.py` gated on the same broad capability and would have re-deleted the DC rows after the platform fix, so it moves to `is_ac_coupled` to match. A useful side-effect: an AIO that lost its DC rows under a prior version gets them re-created at setup, no manual step.

Net behavioural change across every model is **AIO only** — HYBRID, AC single-phase, AC_3PH, EMS, gateway and battery-data-only paths are unchanged.

## One judgement call worth a look

I kept the `and not is_three_phase` guard on the suppression rather than gating on bare `is_ac_coupled`. `AC_3PH` is AC-coupled but gets no AC pair created (the three-phase read-back remaps to different registers — modbus#75), so suppressing its DC pair would leave it with no battery-rate control at all. With the guard, AC_3PH stays a strict no-op. It's the bit I'm least certain about, so worth a second opinion.

The mapping of what HR313/314 actually *does* on an AIO isn't characterised yet — present, holds a value, seemingly inert during charge-rate changes in crg-n's captures — so that's something I'll return to separately.

## Tests

- `test_number.py`: inverted the AIO suppression assertion — the AIO now keeps HR111/112 alongside the AC pair.
- `test_init.py`: added a migration guard that an AIO upgrade retains its DC rows.
- Full suite (603) green; ruff + mypy clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected battery charge and discharge limit controls for All-in-One systems.
  * All-in-One systems now retain DC battery limit controls while also supporting AC-side controls.
  * Refined control visibility for AC-coupled, DC-coupled and three-phase systems.

* **Tests**
  * Added regression coverage to verify correct control retention during upgrades.
  * Updated All-in-One coverage for the expected battery limit controls.

### DIFF
--- a/custom_components/givenergy_local/migrations.py
+++ b/custom_components/givenergy_local/migrations.py
@@ -421,7 +421,7 @@ def _reconcile_per_cell_entities(hass: HomeAssistant, entry: ConfigEntry) -> Non
 def _reconcile_ac_coupled_dc_limits(
     hass: HomeAssistant, coordinator: GivEnergyUpdateCoordinator
 ) -> None:
-    """Remove the DC battery-limit rows on AC-coupled / AIO plants (#52).
+    """Remove the DC battery-limit rows on genuinely AC-coupled plants (#52).
 
     Introduced: v1.3.17 (2026-06). Ongoing while the AC-pair suppression exists
     (cheap no-op once rows are gone).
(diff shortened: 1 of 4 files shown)


dewet22-codex:
Starting review.

dewet22-codex (approved):
Looks good. The split between AC-config surface (`has_ac_config_block`) and AC-coupled control routing (`is_ac_coupled`) matches the upstream capability model, preserves the AIO DC controls, and keeps the registry reconciler aligned with the platform suppression behavior.

Issue #295, opened 2026-07-14
Capture from a Givenergy Hybrid HV Gen 3 (Device Type Code 8102). GIV-HY-8.0-G3-HV


KevC1978:
**What happened?**

Battery SOC remains at 0%

House Consumption remains at 0.00kWh

PV String 1 and 2 Power remains at 0.0w

The inverter supports up to 8000W battery discharge, which is also what the integration reports. However, I only have 3 HV batteries in the stack (10.2kWh), and the GivEnergy Portal shows a maximum battery output of 5990W. I suspect this is due to the lower stack voltage with only 3 batteries installed, resulting in a lower maximum discharge power. 
Increasing the stack size (up to 6 batteries) would likely increase the available voltage and allow the system to reach the full 8000W capability. This might be something to calculate dynamically in future based on battery stack voltage × max discharge current (25A for the 8kW HV inverter, 30A for the 10kW HV inverter), rather than using the inverter's headline power rating.

**Steps to reproduce**

[capture_givenergy_1783981392.txt](https://github.com/user-attachments/files/30010883/capture_givenergy_1783981392.txt)

**Modbus wire capture**
Please attach the capture file you downloaded from this page — drag-and-drop it
into this issue (GitHub doesn't support attaching files via a link). The capture
includes the environment details needed to triage, so there's no need to fill
those in by hand.

dewet22:
Apologies for the slow reply, summer holidays and work got in the way. Thanks for running the three reads, and glad the entities came good. The dumps show every control the integration writes sitting where the single-phase layout expects it, with your actual slot times and target SOC at the right addresses, so the caveat I raised can be downgraded from "reasonably confident" to confirmed once the library folds this bank into its test fixture for your model. I'll close this out when that lands. Your point about deriving max discharge power from stack size stays parked until a second stack size turns up.

PR #302, opened 2026-08-03
feat: adopt givenergy-modbus 2.13.0 (8102 single-phase, PV string V/I) (#281, #295)

dewet22:
Adopts givenergy-modbus 2.13.0. Two upstream fixes land, one of which is a contract change that needs handling on our side.

## `HYBRID_HV_GEN3` is single-phase (#295, modbus#412)

@KevC1978's GIV-HY-8.0-G3-HV (device type 8102) was classified three-phase, so it read the IR(1000–1413) bank — which on that unit answers every read and returns zero in all 414 registers, while the single-phase bank alongside carried real data the whole time. That's why his SOC, House Consumption and PV strings all sat at zero.

Reclassified upstream, so his install decodes properly on this pin: SOC 100%, PV strings 674 W and 4531 W, 21.1 kWh PV today. His entity set will change as the three-phase entities give way to the single-phase ones that actually carry his data.

## `v_pv*` / `i_pv*` return `None` on AC-coupled and All-in-One (#281, modbus#414)

These models have no DC string; the registers carry an AC-side measurement instead. The evidence is the string voltage — on an All-in-One it is *bit-identical* to `v_ac1` across an entire capture, while `i_pv1` sits pinned as power swings (so `p_pv1 != v_pv1 × i_pv1` there, where that product reconciles on every DC hybrid).

Consumer handling here:

- The four sensors become `skip_if_none`, so they stop being created on those models.
- A new `_reconcile_pv_string_vi_sensors` removes the rows an upgraded entry would otherwise keep. Suppressing creation alone leaves four permanently-unknown orphans reporting a mains voltage under a PV-string name — exactly the mislabelling the suppression exists to end.

The reconciler reuses the sensor platform's own gate rather than re-deriving the model logic, so a later upstream widening needs no change here. It guards on a clean seed poll, mirroring `_reconcile_readability_gated_controls`: a partial poll serves last-good, so a `None` read could be a transient bank failure, and removing rows then would destroy a healthy DC hybrid's history.

**The power and energy members are deliberately untouched upstream** — the generation those report is real, just metered AC-side — so the `pv_today`/`pv_power` wiring from #301 is unaffected.

## Known gap, deliberately not papered over

`AC_3PH` keeps a mains-derived `v_pv1`: the upstream routing lives on `SinglePhaseInverter`, and `ThreePhaseInverter` doesn't carry it, so listing that model would look handled while doing nothing. There's no known AC_3PH install today, so it's currently hypothetical. Recorded in a comment so the gap stays visible; upstream want a generating three-phase AC capture before extending it blind.

## Verification

613 pytest, 45 JS. Three new tests: orphan removal on an AC-metered model (verified failing without the reconciler), retention on a partial seed, and retention on a DC hybrid so the suppression is provably model-scoped rather than blanket.

Pin updated in all three places (`pyproject.toml`, `manifest.json`, `uv.lock`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed obsolete PV-string voltage and current sensors when those readings are no longer available.
  * Preserved PV power and energy sensors during this clean-up.
  * Improved handling for partial data updates, preventing valid sensors from being removed prematurely.
  * Ensured PV-string sensors remain available for DC-coupled hybrid systems where applicable.

### DIFF
--- a/custom_components/givenergy_local/__init__.py
+++ b/custom_components/givenergy_local/__init__.py
@@ -79,6 +79,7 @@ from .migrations import (
     _reconcile_per_cell_entities,
     _reconcile_pv_string_vi_sensors,
     _reconcile_readability_gated_controls,
+    _reconcile_reclassified_single_phase_sensors,
     _remove_retired_battery_discharge_year,
 )
 from .predbat import generate_apps_yaml
@@ -648,6 +649,10 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     # current sensors report a mains-derived figure under a PV name; the library now
     # returns None for them and the platform skips them — remove the stale rows (#281).
     _reconcile_pv_string_vi_sensors(hass, coordinator)
+
+    # An 8102 now decodes single-phase rather than three-phase, so the sensors only the
+    # three-phase decode populated are no longer created — remove those rows too (#295).
+    _reconcile_reclassified_single_phase_sensors(hass, coordinator)
 
     # When per-cell exposure is off, remove any individual per-cell rows a prior
     # version (or a prior opt-in) created, so the toggle cleans up rather than
--- a/custom_components/givenergy_local/__init__.py
+++ b/custom_components/givenergy_local/__init__.py
@@ -77,6 +77,7 @@ from .migrations import (
     _reconcile_ems_entities,
     _reconcile_gateway_entities,
     _reconcile_per_cell_entities,
+    _reconcile_pv_string_vi_sensors,
     _reconcile_readability_gated_controls,
     _remove_retired_battery_discharge_year,
 )
@@ -642,6 +643,11 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     # On AC-coupled / AIO plants the DC battery-limit controls are suppressed in
     # favour of the AC pair; remove the stale DC rows an upgraded entry would keep (#52).
     _reconcile_ac_coupled_dc_limits(hass, coordinator)
+
+    # Where PV is metered AC-side there is no DC string, so the per-string voltage and
+    # current sensors report a mains-derived figure under a PV name; the library now
+    # returns None for them and the platform skips them — remove the stale rows (#281).
+    _reconcile_pv_string_vi_sensors(hass, coordinator)
 
     # When per-cell exposure is off, remove any individual per-cell rows a prior
     # version (or a prior opt-in) created, so the toggle cleans up rather than
--- a/custom_components/givenergy_local/manifest.json
+++ b/custom_components/givenergy_local/manifest.json
@@ -13,7 +13,7 @@
   "iot_class": "local_polling",
   "issue_tracker": "https://github.com/dewet22/givenergy-hass/issues",
   "requirements": [
-    "givenergy-modbus==2.12.2"
+    "givenergy-modbus==2.13.0"
   ],
   "version": "1.1.0"
 }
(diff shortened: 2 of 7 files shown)


custom_components/givenergy_local/__init__.py
```diff
@@ -643,6 +644,11 @@ def _system_time_drift_listener() -> None:
     # favour of the AC pair; remove the stale DC rows an upgraded entry would keep (#52).
     _reconcile_ac_coupled_dc_limits(hass, coordinator)
 
+    # Where PV is metered AC-side there is no DC string, so the per-string voltage and
+    # current sensors report a mains-derived figure under a PV name; the library now
+    # returns None for them and the platform skips them — remove the stale rows (#281).
+    _reconcile_pv_string_vi_sensors(hass, coordinator)
```
dewet22-codex, line 651: [P2] Reconcile the sensor rows retired by the 8102 reclassification

This setup reconciliation covers the PV V/I contract change, but the same dependency bump changes existing `HYBRID_HV_GEN3` entries from `ThreePhaseInverter` to `SinglePhaseInverter`. I replayed the linked 8102 capture through both sensor inclusion paths and found seven descriptions included before the bump and excluded afterwards: `battery_maintenance_mode`, `e_battery_charge_total`, `e_battery_discharge_total`, `e_load_total`, `export_power_rate`, `inverter_output_today`, and `inverter_output_total`. Each old value is non-`None` (`battery_maintenance_mode` is `off`; the rest are `0.0`), so existing installations will have registered all seven. HA retains rows when a platform stops adding them, leaving seven permanently unavailable three-phase leftovers after the upgrade. I think the reclassification needs an equivalent clean-seed reconciliation for these old-only keys.

dewet22-codex (changes requested):
I found one upgrade-path issue in the 8102 reclassification. The functional changes and verification are otherwise clean — 613 Python tests, 45 JavaScript tests, Ruff, mypy and lock validation all passed locally — but existing installations retain seven sensor rows that the new single-phase inclusion path no longer creates. I’ve left the exact keys and capture-backed evidence inline.

custom_components/givenergy_local/__init__.py
dewet22-claude: Confirmed and fixed in 3b8dc61. I replayed the 8102 capture through both inclusion paths independently and reproduced your finding exactly — 70 descriptions included under the three-phase decode, 69 after, with the same seven lost. Added `_reconcile_reclassified_single_phase_sensors`, scoped to `HYBRID_HV_GEN3` rather than sweeping every model, since a blanket sweep would delete history wherever a field merely reads None. Keys derive from the platform gate rather than being hardcoded, and it carries the same clean-seed guard as its siblings; two tests, the removal one verified failing without the reconciler.

dewet22-codex (approved):
I re-reviewed the response commit. The new reclassification reconciler is scoped to , derives exclusions through the sensor platform gate, and retains the clean-seed guard. Replaying the 8102 capture still gives the original seven-row old-only delta, and the new tests cover removal and partial-seed retention. I also ran 615 Python tests and 45 JavaScript tests locally, together with Ruff, mypy and lock validation; all passed. The requested change is addressed.

dewet22-codex (approved):
Correction to my preceding approval wording: the new reclassification reconciler is scoped to `HYBRID_HV_GEN3`, derives exclusions through the sensor platform gate, and retains the clean-seed guard. Replaying the 8102 capture still gives the original seven-row old-only delta, and the new tests cover removal and partial-seed retention. I also ran 615 Python tests and 45 JavaScript tests locally, together with Ruff, mypy and lock validation; all passed. The requested change is addressed.

